PR #12, opened 2024-07-07
Update CMakeLists.txt

amontoison:
@cvanaret I updated the main `CMakeLists.txt` such that the users can use `libma57.$dlext` or `libhsl.$dlext`.
It will be also easier for you if in the future you want to support other HSL linear solvers.


cvanaret (approved):

cvanaret:
@amontoison 
Thanks for your contribution, Alexis!

PR #41, opened 2024-10-20
Update documentation in uno.options

odow:
@cvanaret how do I silence all logging? 

Uno is very talkative, which is a bit distracting when we try to solve a large number of problems in sequence.

I tried `ERROR` in https://github.com/jump-dev/AmplNLWriter.jl/actions/runs/11430005564/job/31797031783?pr=190#step:6:123, but it still prints a lot of stuff.

### DIFF
--- a/uno.options
+++ b/uno.options
@@ -20,7 +20,7 @@ print_solution no
 # threshold on objective to declare unbounded NLP
 unbounded_objective_threshold -1e20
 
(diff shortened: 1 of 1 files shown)


cvanaret (approved):

cvanaret:
Good point. I will create a silent logger mode with minimal output.
Thanks for the documentation!

PR #56, opened 2024-10-30
Set print_solution=no in default options

odow:
The options file says no
https://github.com/cvanaret/Uno/blob/950c0fe0285ecb2f4ed3d6c67fad072c43e81856/uno.options#L18

If I solve a problem with a large (e.g., 10^6) number of variables or constraints, this is just going to flood the terminal with numbers. I don't think we want that as a default:

image


cvanaret (approved):

cvanaret:
Indeed. I switched the option when we started the Uno/IPOPT comparison, but `no` by default is more appropriate.

PR #75, opened 2024-11-06
[CI] Compile Uno with HiGHS

amontoison:
Try to compile `Uno_jll.jl` with `HiGHS_jll.jl`.
If it works, we should do the same modification in `.github/julia/build_tarballs_yggdrasil.jl`.


cvanaret (approved):

cvanaret:
It worked :)

Issue #72, opened 2024-11-04
Interface to HiGHS as an LP/QP solver


cvanaret:
Add an interface to [HiGHS](https://ergo-code.github.io/HiGHS/stable/) as an LP solver and convex QP solver.
This should be relatively simple, since HiGHS is implemented in C++.

cvanaret:
An interface to the LP solver HiGHS was added in https://github.com/cvanaret/Uno/pull/74.

For QPs, it's a bit trickier because HiGHS expects a compressed sparse Hessian in one of the following shapes:
- column-wise lower triangular;
- row-wise upper triangular.

At the moment, the CSC sparse format in Uno is only column-wise upper triangular. I need to investigate this further.

amontoison:
@cvanaret 
I did this Julia code a few months ago to get the sparse CSC format of A' based on the CSC format of A.
It could be what you need:
```julia
function MyTranspose(A::SparseMatrixCSC{T,Cint}) where {T}
    m, n = size(A)
    nnzA = nnz(A)
    A_colptr = A.colptr
    A_rowval = A.rowval
    A_nzval = A.nzval

    # Allocate storage for the column pointers and row indices of B = Aᵀ
    B_colptr = zeros(Cint, m + 1)
    B_rowval = Vector{Cint}(undef, nnzA)
    B_nzval = Vector{T}(undef, nnzA)


    # Count the number of non-zeros for each row of A.
    # It corresponds to the number of non-zeros for each column of B = Aᵀ.
    for k in 1:nnzA
        i = A_rowval[k]
        B_colptr[i] += 1
    end

    # Compute the cumulative sum to determine the starting positions of rows in B_rowval
    counter = 1
    for col in 1:m
        nnz_col = B_colptr[col]
        B_colptr[col] = counter
        counter += nnz_col
    end
    B_colptr[m + 1] = counter

    # Store the row indices for each column of B = Aᵀ
    for j in 1:n
        for index in A_colptr[j]:(A_colptr[j + 1] - 1)
            i = A_rowval[index]

            # Update B_rowval for the non-zero B[j,i].
            # It corresponds to the non-zero A[i,j].
            pos = B_colptr[i]
            B_rowval[pos] = j
            B_nzval[pos] = A_nzval[index]
            B_colptr[i] += 1
        end
    end

    # Fix offsets of B_colptr to restore correct starting positions
    for col in m:-1:2
        B_colptr[col] = B_colptr[col - 1]
    end
[...]

PR #77, opened 2024-11-07
Enable LP subproblem in julia/runtests.jl

odow:
x-ref #72

### DIFF
--- a/.github/julia/runtests.jl
+++ b/.github/julia/runtests.jl
@@ -60,6 +60,8 @@ Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
             "006_010",
             # Remove once https://github.com/cvanaret/Uno/issues/38 is fixed
             "007_010",
+            # TODO: investigate why this fails
+            "008_010",
         ],
         primal_target,
         objective_tol = 1e-4,
@@ -74,6 +76,12 @@ Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
         primal_target,
         objective_tol = 1e-4,
         primal_tol = 1e-4,
+        exclude = [
+            # TODO: investigate why this fails
+            "501_010",
+            # TODO: investigate why this fails
+            "501_011",
+        ],
     )
 end
 
--- a/.github/julia/runtests.jl
+++ b/.github/julia/runtests.jl
@@ -22,7 +22,7 @@ function Optimizer(options = String["logger=SILENT"])
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
-Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
+Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp", "max_iterations=10000"])
 
 # This testset runs https://github.com/jump-dev/MINLPTests.jl
 @testset "MINLPTests" begin
@@ -53,6 +53,7 @@ Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
         exclude = [
             "001_010",  # Local solution
             "003_014",  # Local solution
+            "008_010",  # Local solution
             # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
             "005_010",
             # Okay to exclude forever: AmplNLWriter does not support
@@ -60,8 +61,6 @@ Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
             "006_010",
             # Remove once https://github.com/cvanaret/Uno/issues/38 is fixed
             "007_010",
-            # TODO: investigate why this fails
-            "008_010",
         ],
         primal_target,
         objective_tol = 1e-4,
@@ -76,12 +75,6 @@ Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
         primal_target,
         objective_tol = 1e-4,
         primal_tol = 1e-4,
-        exclude = [
-            # TODO: investigate why this fails
-            "501_010",
-            # TODO: investigate why this fails
-            "501_011",
-        ],
     )
 end
 
--- a/.github/julia/runtests.jl
+++ b/.github/julia/runtests.jl
@@ -18,11 +18,11 @@ import Uno_jll
 Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
 solver.
 """
-function Optimizer(options = String["logger=INFO"])
+function Optimizer(options = String["logger=SILENT"])
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
-Optimizer_LP() = Optimizer(["logger=INFO", "preset=filterslp"])
+Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
 
 # This testset runs https://github.com/jump-dev/MINLPTests.jl
 @testset "MINLPTests" begin
--- a/.github/julia/runtests.jl
+++ b/.github/julia/runtests.jl
@@ -35,22 +35,46 @@ Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
     # This function tests (potentially) non-convex nonlinear programs. The tests
     # are meant to be "easy" in the sense that most NLP solvers can find the
     # same global minimum, but a test failure can sometimes be allowed.
-    exclude = [
-        # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
-        "005_010",
-        # Okay to exclude forever: AmplNLWriter does not support
-        # user-defined functions.
-        "006_010",
-        # Remove once https://github.com/cvanaret/Uno/issues/38 is fixed
-        "007_010",
-    ]
-    MINLPTests.test_nlp_expr(Optimizer; exclude, primal_target)
-    MINLPTests.test_nlp_expr(Optimizer_LP; exclude, primal_target)
+    MINLPTests.test_nlp_expr(
+        Optimizer;
+        exclude = [
+            # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
+            "005_010",
+            # Okay to exclude forever: AmplNLWriter does not support
+            # user-defined functions.
+            "006_010",
+            # Remove once https://github.com/cvanaret/Uno/issues/38 is fixed
+            "007_010",
+        ],
+        primal_target,
+    )
+    MINLPTests.test_nlp_expr(
+        Optimizer_LP;
+        exclude = [
+            "001_010",  # Local solution
+            "003_014",  # Local solution
+            # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
+            "005_010",
+            # Okay to exclude forever: AmplNLWriter does not support
+            # user-defined functions.
+            "006_010",
+            # Remove once https://github.com/cvanaret/Uno/issues/38 is fixed
+            "007_010",
+        ],
+        primal_target,
+        objective_tol = 1e-4,
+        primal_tol = 1e-4,
+    )
     # This function tests convex nonlinear programs. Test failures here should
     # never be allowed, because even local NLP solvers should find the global
     # optimum.
     MINLPTests.test_nlp_cvx_expr(Optimizer; primal_target)
-    MINLPTests.test_nlp_cvx_expr(Optimizer_LP; primal_target)
+    MINLPTests.test_nlp_cvx_expr(
+        Optimizer_LP; 
+        primal_target,
+        objective_tol = 1e-4,
+        primal_tol = 1e-4,
+    )
 end
 
 # This testset runs the full gamut of MOI.Test.runtests. There are a number of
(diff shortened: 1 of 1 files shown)


odow:
So I guess we need some other options? 

https://github.com/cvanaret/Uno/actions/runs/11731547352/job/32681840126?pr=77#step:8:23697

cvanaret:
I fixed some stuff in #82.

You can then try `preset=filtersqp` and `subproblem=LP`, which will overwrite the subproblem strategy in the `filtersqp` preset. We still may have troubles with infeasible instances, I need to fix something there.

.github/julia/runtests.jl
```diff
@@ -18,11 +18,12 @@ import Uno_jll
 Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
 solver.
 """
-function Optimizer()
-    options = String["logger=INFO"]
+function Optimizer(options = String["logger=INFO"])
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
+Optimizer_LP() = Optimizer(["logger=INFO", "subproblem=LP"])
```
odow:
```suggestion
Optimizer_LP() = Optimizer(["logger=INFO", "preset=filtersqp", "subproblem=LP"])
```

.github/julia/runtests.jl
```diff
@@ -18,11 +18,12 @@ import Uno_jll
 Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
 solver.
 """
-function Optimizer()
-    options = String["logger=INFO"]
+function Optimizer(options = String["logger=INFO"])
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
+Optimizer_LP() = Optimizer(["logger=INFO", "preset=filtersqp", "subproblem=LP"])
```
odow: @cvanaret I get the error `The option with name QP_solver was not found`

cvanaret: @odow That's weird. I cannot reproduce the error with the command line. Is it possible that AmplNLWriter passes the options in a different order?

odow: > Is it possible that AmplNLWriter passes the options in a different order?

Yes. We store options in a `Dict`, which has an arbitrary iteration order:
https://github.com/jump-dev/AmplNLWriter.jl/blob/1141925e427915be13f45b8b53e3c6415f782e8a/src/AmplNLWriter.jl#L104

Note that JuMP does not have the concept of "ordered" options. I'd classify this as a bug in Uno. 

At minimum, users having to know that the behaviour of Uno depends on the options _and the order in which they appear_ is a massive footgun that is going to cause issues down the track.

cvanaret: Ah. In the case of Uno, that's annoying. Precedence lifts any ambiguity in situations like `preset=ipopt option_file=uno.options` and is very useful to overwrite a few options from a preset like we're trying to do here. I think it's very expressive.
What would be your suggestion to address this?

cvanaret: In the meantime, you can run the tests with the following options:
```julia
Optimizer_LP() = Optimizer(["LP_solver=HiGHS", "TR_min_radius=1e-8", "TR_radius=10",
"constraint_relaxation_strategy=feasibility_restoration", "filter_type=standard", "globalization_mechanism=TR",
"globalization_strategy=fletcher_filter_method", "l1_constraint_violation_coefficient=1.", "logger=INFO",
"loose_tolerance=1e-6", "progress_norm=L1", "protect_actual_reduction_against_roundoff=no", "residual_norm=L2",
"sparse_format=CSC", "subproblem=LP", "switch_to_optimality_requires_linearized_feasibility=yes",
"tolerance=1e-6"])
```

odow: cc @blegat thoughts?

> What would be your suggestion to address this?

You define `preset` to be processed first. Then any other options override. I don't see the ambiguity with `preset` and `option_file`.

You also don't check option validity (does `QP_solver` exist, etc) until all options have been processed.

odow: I don't think we have an example of any other solver which has behaviour that depends on the order of the options.

odow:
@cvanaret have you tried to solve problems with HiGHS locally? You might want to cancel the CI run, because it seems to be taking forever.

odow:
The log is full of runs that look like:

image

and

image

cvanaret:
> @cvanaret have you tried to solve problems with HiGHS locally?

Yes, I ran the SLP method with HiGHS on our subset of 429 CUTEst instances (it can solve around 60% - that's the curse of subproblems without curvature...) as well as a few infeasible instances (all solved).
I will add a couple of instances as unit tests to avoid cluttering up the CI.

> The log is full of runs that look like:

Can you post the JuMP models of these two instances?

odow:
I just picked two random screenshots. Here are a few that failed tests. The cvx ones are probably good to start with, because they should have a unique global solution.

https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-cvx-expr/001_011.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-cvx-expr/002_011.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-cvx-expr/101_010.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-cvx-expr/102_012.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-cvx-expr/102_013.jl

https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/001_010.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/002_010.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/003_014.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/003_015.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/008_010.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/009_010.jl
https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/009_011.jl

.github/julia/runtests.jl
blegat: We have [SDPA](https://github.com/jump-dev/SDPA.jl/) which has something similar. There are three Modes that give good set of values for parameters but you can also individually set all parameters.
My understanding is that SDPA first process the modes and then set the other parameters. It seems like a good approach that could also be used for the `preset`. If you specifically set a parameter, it seems intuitive that it will take precendence over a `preset`. So the order is done on what's more specific has priority and not the last parameter set has priority.

cvanaret:
@odow Thanks!
Locally, my trust-region SLP + HiGHS converges on all instances and gets the same result as my trust-region SQP + BQPD (except for `nlp-expr/001_010` but the respective minimizers are both global, see https://github.com/jump-dev/MINLPTests.jl/issues/50). For the maximization problem `nlp-expr/003_015.jl`, both methods find the local maximum (0, 2).

PRs #83 and #84 should have fixed a lot of issues.

.github/julia/runtests.jl
cvanaret: @odow @blegat 
I understand and I realize that this behavior is atypical for a solver. This comes from the fact that there are three possibilities to set options: preset, (partial) option file and individual option.

If a particular option (say `LP_solver`) were to appear in all three possibilities simultaneously, we'd need a way to establish precedence to figure out its actual value. I just figured that the command-line order is more intuitive and predictable that having an implicit order that users have to know about and comply with.

That said, if you think this is an absolute dealbreaker, I'll make the change ;)

cvanaret:
@odow at least this was a quick fail :D
The problem here is that it's taking the default preset (`ipopt`) and replacing the subproblem with `LP`. Therefore it's not using a trust region and subproblems get unbounded. Can you try with the long list of options I wrote in the other message?

odow:
Ugh. When I rebased on your latest stuff I dropped that commit :cry:

cvanaret:
OK looking at the first fails in the log:
- `/tmp/jl_PWcf5p/model.nl`: fail is documented in https://github.com/jump-dev/MINLPTests.jl/issues/50.
- `/tmp/jl_NYSwin/model.nl`: local maximum.
- `/tmp/jl_buq6o8/model.nl` and `/tmp/jl_SZ6RRC/model.nl`: cannot reduce stationarity (I guess the duals don't settle fast enough). We may have to reduce the tolerance (or find out if there's a bug).

I will look at the complete list and try and figure out what went wrong.

.github/julia/runtests.jl
blegat: The configuration file indeed makes it a bit more tricky. If you have an option set both in the config file and explicitly, it's less clear what to do. One option is to error in that case. Otherwise, you could say that the preset is set first, then the option file and then the individual options. If the user sets an option explicitly, my intuition is that it takes precedence over file and preset, it might be confusing if this depends on the order of the arguments.

odow: Parse it in this order

1. Option file
2. cmd options
   a. preset
   b. all others 

or just splat the option file into the list of options, and error if something is set twice.

I don't really care. Just pick a convention, document it, and move on.

> That said, if you think this is an absolute dealbreaker, I'll make the change ;)

This is an absolute deal breaker. The solver behaviour should not depend on the order in which options are passed.

odow:
> cannot reduce stationarity (I guess the duals don't settle fast enough). We may have to reduce the tolerance

Open to different parameter settings

.github/julia/runtests.jl
```diff
@@ -18,11 +18,32 @@ import Uno_jll
...
+        "LP_solver=HiGHS",
+        "TR_min_radius=1e-8",
+        "TR_radius=10",
+        "constraint_relaxation_strategy=feasibility_restoration",
+        "filter_type=standard",
+        "globalization_mechanism=TR",
+        "globalization_strategy=fletcher_filter_method",
+        "l1_constraint_violation_coefficient=1.",
+        "logger=INFO",
+        "loose_tolerance=1e-6",
```
odow:
```suggestion
        "loose_tolerance=1e-4",
```

.github/julia/runtests.jl
```diff
@@ -18,11 +18,32 @@ import Uno_jll
...
+        "progress_norm=L1",
+        "protect_actual_reduction_against_roundoff=no",
+        "residual_norm=L2",
+        "sparse_format=CSC",
+        "subproblem=LP",
+        "switch_to_optimality_requires_linearized_feasibility=yes",
+        "tolerance=1e-6",
```
odow:
```suggestion
        "tolerance=1e-5",
```

.github/julia/runtests.jl
cvanaret: Fixed in https://github.com/cvanaret/Uno/pull/85.

cvanaret:
@odow I implemented the precedence rules for setting options (#85) and I added a `filterslp` preset (trust-region filter SLP) with similar options to `filtersqp` (#88).
Can you rebase and replace the option list with simply `preset=filterslp`? This should pass more tests.

.github/julia/runtests.jl
odow:
```suggestion
Optimizer_LP() = Optimizer(["logger=INFO", "preset=filterslp"])
```

.github/julia/runtests.jl
```diff
@@ -18,11 +18,12 @@ import Uno_jll
 Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
 solver.
 """
-function Optimizer()
-    options = String["logger=INFO"]
+function Optimizer(options = String["logger=INFO"])
```
odow:
```suggestion
function Optimizer(options = String["logger=SILENT"])
```

.github/julia/runtests.jl
```diff
@@ -18,11 +18,12 @@ import Uno_jll
...
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
+Optimizer_LP() = Optimizer(["logger=INFO", "preset=filterslp"])
```
odow:
```suggestion
Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp"])
```

cvanaret:
Thanks for looking at the failures. It was a good call to relax the tolerance of the tests.

Here's some feedback:
- `nlp_expr/008_010`: local solution.
- `nlp_cvx_expr/501_011`: I haven't tested all 20 dimensions, but the largest instance (n = 20) is solved locally in 3248 iterations (SLP methods are just sloooow), which is over the default limit 2000. So perhaps we can bump it (`max_iterations=10000`).
- `nlp_cvx_expr/501_010`: same as `nlp_cvx_expr/501_011` but requires fewer iterations.

odow:
Once CI finishes this is good to go from my end.

cvanaret:
It's a wrap. Thanks a lot!

cvanaret (approved):

odow:
Merge #78 and then tag Uno@1.3.0?

cvanaret:
[Uno 1.3.0](https://github.com/cvanaret/Uno/releases/tag/v1.3.0) is out!

PR #83, opened 2024-11-11
TR strategy: properly reset the bounds duals and set the warm-start information

cvanaret:
Properly update the WarmstartInformation object: handle phase switches, errors, etc.
Two fixes in TrustRegionStrategy:
- loop over all variables to reset the multipliers when TR constraints are active;
- reset the feasibility multipliers as well;

### DIFF
--- a/uno/Uno.cpp
+++ b/uno/Uno.cpp
@@ -11,6 +11,7 @@
 #include "linear_algebra/Vector.hpp"
 #include "model/Model.hpp"
 #include "optimization/Iterate.hpp"
+#include "optimization/WarmstartInformation.hpp"
 #include "solvers/QPSolverFactory.hpp"
 #include "solvers/LPSolverFactory.hpp"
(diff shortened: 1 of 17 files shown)


odow:
Can we test the TR without BQPD?

cvanaret:
Yes, we can assemble a trust-region SLP method with HiGHS as LP solver.
So far, I've fixed lack of convergence on infeasible problems for SLP methods. I'm still working on properly setting the warmstart information.

leyffer:
Excellent.
It would be good to also test it with BQPD, just to make make sure the two solvers behave the same on smaller instances.
Eventually, we should add an EQP phase and get a second order method.
Sven

Sven Leyffer
Senior Computational Mathematician and Deputy Director
Mathematics and Computer Science Division
Argonne National Laboratory
________________________________

odow:
> It would be good to also test it with BQPD

Sure. But this requires us to be able to compile a version of Uno that includes BQPD on a public CI instance.

Issue #106, opened 2024-11-19
Add user callbacks


cvanaret:
We should allow a few user-defined callbacks:
- when an iterate is accepted;
- when a new primal iterate is found;
- when a new dual iterate is found;
- etc

Reported by @worc4021.

PR #108, opened 2024-11-20
Suggestion for implementing user callbacks

cvanaret:
Pass an object `UserCallbacks` around and call `notify_*` member functions upon finding new acceptable iterate, new primals and new multipliers (the last two are a redundant with the first one - to be discussed).

@worc4021 would that work for you?

Fixes https://github.com/cvanaret/Uno/issues/106.


worc4021:
Looks perfect to me!

PR #115, opened 2024-11-30
MA27 interface

worc4021:
This PR adds a ma27 linear solver interface. I expect that the existing build should already work with this.

A mechanism for resizing the work arrays used by ma27 is provided. A unit test equivalent to the MA57 is also provided and passes. 

However, I have observed that when attempting to solve hs15 with the byrd preset the problem is repeatedly rank deficient and eventually the nlp fails due to inertia correction becoming unstable. Unclear, what happens there. The ipopt preset is happy with it and other test problems I tried didn't trigger this behaviour. One cause for this might be that ma27 and ma57 differ in an important aspect: ma27 stores the factors in the array passed with the values of A whereas ma57 has a separate store for it. Hence, in the existing ma57 interface it is possible to pass `matrix.data_pointer()` directly to the factorisation and solver routine, whereas in ma27 we have to make a copy and keep it for subsequent calls. The documentation states that the solve should not be called with different values of `A` from the factorisation, but in principle it would be possible to pass different values. Not sure if this is done though.


cvanaret:
Thanks @worc4021! I will investigate the issue you described with the byrd preset and hs015. We may have to change some settings to better estimate the inertia.

PR #137, opened 2024-12-12
[WIP] Upgrade from 4 to 8 ingredients

cvanaret:
The goal of this PR is to upgrade from 4 to 8 ingredients:

- reformulation layer:
   - constraint relaxation strategy
   - inequality handling method
 - subproblem layer:
   - local model
   - hessian model
   - regularization strategy
 - subproblem solver layer
   - subproblem solver
 - globalization layer:
   - globalization strategy
   - globalization mechanism.

Steps:
- [x] introduce a `LagrangeNewtonSubproblem` ingredient
- [x] introduce `InequalityHandlingMethod` ingredient
- [x] introduce `HessianModel` ingredient
- [x] introduce `RegularizationStrategy` ingredient
- [x] pass `LagrangeNewtonSubproblem` to QP/LP subproblem solvers
- [x] pass `LagrangeNewtonSubproblem` to linear solvers
- [x] defer the Hessian computation of the `Model` to the `HessianModel`
- [ ] pass `LagrangeNewtonSubproblem` to globalization strategy
- [x] rename `QPSolver` and `SymmetricLinearSolver` into `InequalityQPSolver` and `EqualityQPSolver`
- [x] merge classes `QPMethod` and `LPMethod` into class `InequalityConstrainedMethod`
- [ ] handle `InequalityQPSolver` and `LPSolver` in `InequalityConstrainedMethod`
- [ ] fix computation of least-square multipliers
- [ ] allocate and store the linear solver required in `PrimalRegularization` and `PrimalDualRegularization`
- [x] allocate memory for matrices only in subproblem solvers
- [ ] create matrix views in subproblem solvers for evaluation of Hessian and Jacobian
- [x] compute number of Hessian nonzeros for memory allocation
- [ ] restore unused `ScaledModel`

PR #145, opened 2025-03-12
Github action to build with MSVC

cvanaret:
Github action to build with MSVC

Issue #144, opened 2025-03-12
Windows MSVC support?


rujialiu:
Dear sir,

First of all, thank you for this awesome solver! I tried to compile Uno with MSVC 2022 on windows. There are a few errors but I managed to solve them. There aren't many changes:

* MSVC doesn't support C++ keyword `not` so I need to change them to `!`
* There is one error regarding templated function signature matching. Just a few extra characters would make it compile
* A few source files contain non-ascii characters. MSVC needs a pragma to explicitly tell the compile it's in UTF-8, and the file needs BOM (I don' t know whether there are less `intrusive` ways to resolve this without adding BOM)
* The current cmake config contains some gcc-specific things that need 

I wonder if you're willing to support Windows/MSVC. If so, I'll make a PR.

cvanaret:
Dear @rujialiu,

Thank you for the kind words!
Please, do make a PR. I will also add MSVC to our CI scripts (at the moment we only build with GCC on Windows).
Best regards,

Charlie

cvanaret:
Closed with https://github.com/cvanaret/Uno/pull/146

PR #146, opened 2025-03-12
Build Uno with MSVC on windows

rujialiu:
I don't have gcc installed on my machine so I didn't check whether gcc is broken (the only possible broken is the templated function). Hope CI will tell us the result :)

### DIFF
--- a/.github/workflows/build-msvc.yml
+++ b/.github/workflows/build-msvc.yml
@@ -1,0 +1,50 @@
+# This starter workflow is for a CMake project running on a single platform. There is a different starter workflow if you need cross-platform coverage.
+# See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-multi-platform.yml
+name: Build with MSVC
+
+on:
+  push:
+    branches: [ "main" ]
+    paths-ignore:
+      - '*.md'
+      - 'LICENSE'
+      - '*.cff'
+      - '*.options'
+      - 'uno_ampl-completion.bash'
+  pull_request:
+    branches: [ "main" ]
+    paths-ignore:
+      - '*.md'
+      - 'LICENSE'
+      - '*.cff'
+      - '*.options'
+      - 'uno_ampl-completion.bash'
+    
+jobs:
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally well on Windows or Mac.
+    # You can convert this to a matrix build if you need cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [windows-latest]
+        architecture: [x64, x86]
+        # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+        BUILD_TYPE: [Release, Debug]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up environment to compile C++ with MSVC
+      uses: ilammy/msvc-dev-cmd@v1
+
+    - name: Configure CMake
+      # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
+      # See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type
+      run: cmake -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=${{matrix.BUILD_TYPE}} .
+
+    - name: Build
+      # Build your program with the given configuration
+      run: nmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,14 @@ project(Uno VERSION 1.3.0
         DESCRIPTION "Uno (Unifying Nonconvex Optimization)"
         LANGUAGES C CXX)
 set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wnon-virtual-dtor -pedantic -Wunused-value -Wconversion")
-set(CMAKE_CXX_FLAGS_DEBUG "-pg")
-set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG") # disable asserts
+if (MSVC)
+   add_compile_options("$<$<COMPILE_LANGUAGE:C>:/utf-8>")
+   add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:/utf-8>")
+else()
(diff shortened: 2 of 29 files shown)


rujialiu:
Addresses #144 @cvanaret

rujialiu:
Note that the CMakeLists.txt is not exactly the same one I used, because I don't have Fortran compiler installed. It might be better to disable fortran if the user doesn't have it and doesn't need Fortran FFI, but I'm too lazy to change CMakeLists.txt (I just remove Fortran related things locally).

Also, just curious: by skimming over the codes, if I have HiGHS and wants to use filterslp only, it looks like even blas/lapack is optional? So the only real hard dependency is ASL (but in INSTALL.md BLAS/LAPACK doesn't look like optional)? I'm developing a small tool that solves some easy nonlinear programming instances so I want to make it as lightweight as possible (that's why I didn't consider anything that requires Ipopt).

cvanaret:
@rujialiu 
Great, nothing got broken :) Thanks! Now we need to add a CI script to build with MSVC.

I agree, we should get rid of the Fortran FFI if it is not needed (this will also help with the aforementioned CI script).
I could rewrite the CMake as follows:
- [x] first detect the third-party libraries;
- [x] set a flag `REQUIRES_FORTRAN` if the HSL or BQPD libraries are present;
- [x] if `REQUIRES_FORTRAN` is true, add the language `Fortran` to the project and include `FortranCInterface` .

Let me know whether that would work.

Yes, it looks like BLAS and LAPACK aren't required if you only need HiGHS. I will update the README.

rujialiu:
> Let me know whether that would work.
Yes, that looks great :)

> Yes, it looks like BLAS and LAPACK are optional if you only need HiGHS. I will update the README.
That's great!

cvanaret:
The Fortran FFI is properly enabled in https://github.com/cvanaret/Uno/pull/147.

cvanaret:
@rujialiu
I merged #147 to the main branch. Can you:
- rebase your branch;
- add a Github action like this one? https://github.com/cvanaret/Uno/pull/145/files#diff-25938a76bb4ebc1a13b511013f09be467a36a4e66c3425efb6f12507cdd1d4ef

rujialiu:
Ok! Do you want me to merge that branch or manually commit the CI file?

cvanaret:
Yeah, you can manually commit it.

cvanaret:
Works perfectly :) Last thing to add would be your name to the list of contributors in the README.

rujialiu:
> Works perfectly :) Last thing to add would be your name to the list of contributors in the README.

Done! I hope I'll be able to make some more complex contributions in the future :)

rujialiu:
BTW: The contributor list is already out of alphabetic order so I just added my self to the end of list :)

cvanaret:
Thanks a lot! Give me some feedback when you get it to run ;)

PS: the family names are alphabetically sorted.

rujialiu:
> Thanks a lot! Give me some feedback when you get it to run ;)
It solves hs015.nl sucessfully and crashed on polak5 (after several iterations) :-S
Since I'm very unfamiliar with Uno and OR in general, I need some time to understand what's the root cause. The direct cause is here:

```c++
// complementary slackness error: expression for violated constraints depends on the definition of the relaxed problem
double l1RelaxedProblem::complementarity_error(const Vector<double>& primals, const std::vector<double>& constraints,
const Multipliers& multipliers, double shift_value, Norm residual_norm) const {
    // bound constraints
    const Range variables_range = Range(this->number_variables);
    const VectorExpression bounds_complementarity{variables_range, [&](size_t variable_index) {
        if (0. < multipliers.lower_bounds[variable_index]) {
            return multipliers.lower_bounds[variable_index] * (primals[variable_index] - this->variable_lower_bound(variable_index)) - shift_value; // <--- crashed here. variable_index=3, primals has 3 values
        }
```

`this->number_variables` is 5 so later `bounds_complementarity` is called with `variable_idx=3` but `primals` only has 3 variables. But since I can see it's some kind of auxiliary problems maybe it really has 5 variables but `primals` fails to expand its size? I don't know. Anyway, `this->number_variables` is 5 from the first breakpoint hit, but in early iterations it never got into the `if` statement so it didn't crash earlier.

I suspect it has something to do with ASL because the code looks ancient with lots of workarounds/tricks. I understand it's written to support old compilers but it's really easy to get wrong. I even had to disable Control87 otherwise there will be assertion error on startup (even before the file is read).

It'll be great if we can use Uno without ASL, e.g. just pass a function to calculate value and first/second order derivatives. It can be easily done with things like `autodiff`(https://github.com/autodiff/autodiff). Actually I had very positive experience with nlopt which accepts callbacks as I said. I thought (because I don't know much about OR) ASL can read `.mod` files directly, but after I successfully built Uno + ASL I realized it can only read `.nl`. I had to use AMPL to convert `.mod` to `.nl` which adds another dependency (is there any more lightweight way to do `.mod`->`.nl` conversion?). So using Uno with callback seems to be my preferred way. If you have this on our to-do list but not planned in near future I can try to contribute :)

> PS: the family names are alphabetically sorted.
Oops... careless me.

cvanaret:
Oh that's nasty! I can reproduce the error when I add an `assert()` in the loop. It systematically crashes, not just for polak5 😨 Moral of the story: I should use more assertions and spend more time in Debug mode.
Can you open a Github issue, copy-paste the Uno output and mention the corresponding PR https://github.com/cvanaret/Uno/pull/149?

I haven't heard of another way to convert `.mod` to `.nl`. I agree that the dependency of Uno of AMPL/ASL is annoying (even though it has the abstraction to work with other models).

Using Uno with function pointers/callbacks is definitely on my to-do list. When the massive code upgrade I'm working on is finally done, we'll have a more stable `Model` interface and I'll be able to make progress on that. I'll keep you posted :)

rujialiu:
Wow you are so fast! Maybe you can add some local scripts to run uno with ASan enabled. Glad to hear you have an  ongoing big refactor. Would love to alpha test it whenever possible

cvanaret:
I tried to [compile/link with ASan](https://felsoci.sk/blog/using-address-sanitizer-asan-in-a-cmake-project.html), but to no avail :( The solver just runs normally.

You can follow the big code upgrade here: https://github.com/cvanaret/Uno/pull/137

rujialiu:
> You can follow the big code upgrade here: #137

Gently ping @cvanaret 
Now I see #137 is closed without merging, does that mean the big code upgrade was failed? But since version 2 is released, I think maybe it's done by other work?

> Using Uno with function pointers/callbacks is definitely on my to-do list. When the massive code upgrade I'm working on is finally done, we'll have a more stable Model interface and I'll be able to make progress on that. I'll keep you posted :)
Any progress on this? I hope I can help with testing even before it's done.

cvanaret:
Hi @rujialiu ,
I closed #137 and merged the changes incrementally in subsequent PRs. The resulting refactoring was released as part of v2.0.0.
I'm planning to write a C interface in the next few weeks :)

rujialiu:
Nice to hear that! I can do some early testing if you need some feedback before finalizing the interface. @cvanaret

cvanaret:
Thanks! I'll tag you when I open the PR.

PR #168, opened 2025-06-07
Cross-compile Uno with BQPD

amontoison:
`BQPD_jll.jl` is an official Julia package now and we can cross-compile Uno with it.

### DIFF
--- a/.github/julia/runtests.jl
+++ b/.github/julia/runtests.jl
@@ -18,11 +18,11 @@ import Uno_jll
 Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
 solver.
 """
-function Optimizer(options = String["logger=SILENT", "unbounded_objective_threshold=-1e15"])
+function Optimizer(options = String["logger=SILENT", "preset=ipopt", "unbounded_objective_threshold=-1e15"])
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
-Optimizer_LP() = Optimizer(["logger=SILENT", "preset=ipopt", "LP_solver=HiGHS", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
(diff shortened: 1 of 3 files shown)


cvanaret:
@amontoison that's great news!!
Some MINLP tests failed, I believe because we replaced the LP solver HiGHS with BQPD in the `filterslp` preset, and they sometimes return different solutions.
Could you try adding the option `LP_solver=HiGHS` [here](https://github.com/cvanaret/Uno/blob/main/.github/julia/runtests.jl#L25)?

cvanaret:
@amontoison OK I cracked it. Now that BQPD is available, the `filtersqp` preset is now default. Can you add `preset=ipopt` [here](https://github.com/cvanaret/Uno/blob/main/.github/julia/runtests.jl#L21)?

.github/julia/runtests.jl
```diff
@@ -22,7 +22,7 @@ function Optimizer(options = String["logger=SILENT", "unbounded_objective_thresh
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
-Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
+Optimizer_LP() = Optimizer(["logger=SILENT", "preset=filterslp", "LP_solver=HiGHS", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
```
amontoison, line 25:
```suggestion
Optimizer_LP() = Optimizer(["logger=SILENT", "preset=ipopt", "LP_solver=HiGHS", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
```

cvanaret:
@amontoison I meant on line 21, not 25 ;) I pushed the fix.

amontoison:
I was in the Uber :)
On my way to the airport for going to the Householder symposium.

cvanaret:
Ahah no worries!
We're good to go, thanks Alexis :)

PR #190, opened 2025-06-25
[CI] Compile Uno with MUMPS

amontoison:
I cross-compiled MUMPS 5.8.0 (static mode) with METIS + reference BLAS  / LAPACK.
-> https://github.com/amontoison/MUMPS_static_jll.jl


cvanaret:
Thanks Alexis!
So there seem to be two issues:
- BQPD is somehow not found by the macOS build.
- my logic in `QPSolverFactory` is flawed, it won't compile when no QP solver is available 😅
I'll check the CI script.

cvanaret (commented):
Somehow some arguments of CMake are ignored. Investigating

cvanaret:
@amontoison
METIS cannot be found for macOS. Can you set the `METIS_LIBRARY` path?

amontoison:
Yes, I can do that :)
I am working in parallel with Iain Duff in Glasgow, do that asap.

cvanaret:
@amontoison 
Perfect! The only issue remaining is in `unit-tests-ubuntu.yml`: BLAS and LAPACK don't seem to be installed. Perhaps we can load an action there.

amontoison:
`libblas.a` and `liblapack.a` are in the archive with `libmetis.a` and `libdmumps.a` so the issue is how you link it. ;)

amontoison:
@cvanaret 
Capture d’écran du 2025-06-26 13-44-47

cvanaret:
Indeed!
Here's the CMake configuration:
```
-- Found BLAS: ./deps/lib/libblas.a
-- Found LAPACK: ./deps/lib/liblapack.a
```
then linking:
```
/usr/bin/ld: cannot find -l./deps/lib/libblas.a: No such file or directory
/usr/bin/ld: cannot find -l./deps/lib/liblapack.a: No such file or directory
```

amontoison:
I think that the linking is done from a different folder that the "root" of Uno and the relative path is not valid anymore.
The question is why does it work for MUMPS and BQPD and not for BLAS / LAPACK?

amontoison:
@cvanaret I added absolute paths everywhere and it fixed the issue. You should probably add a comment about it somewhere.
The current error is that you don't correctly link with BLAS / LAPACK in Uno.
If we have two libraries (one for BLAS and one for LAPACK), it seems that you only link to `liblapack.a` and not `libblas.a`, which is required by LAPACK.

amontoison:
@cvanaret All tests passed, I fixed a bug in the `CMakeLists.txt`, we must link BLAS after LAPACK if they are static libraries.

cvanaret:
Wonderful, thank you for looking into this!

PR #198, opened 2025-07-02
Deleted BQPD's alphac definition

cvanaret:
v2.0.0 release failed because of BQPD's alphac. Removed definition altogether.


amontoison:
If the symbol is not defined in BQPD, it will not help :)

cvanaret:
BQPD has several common blocks (one of them being `alphac`), and it looks like if you don't need to set the values of a block, you don't even have to define it in the C++ code.

amontoison (approved):

PR #199, opened 2025-07-02
Fix the cross-compilation with BQPD

amontoison:
I expect that `libbqpd.a` is included in `libuno.a` but if we have again the issue to cross-compile UNO v2.0, this PR should definitively fix the problem.

Please do not merge if not needed.

### DIFF
--- a/.github/julia/build_tarballs_release.jl
+++ b/.github/julia/build_tarballs_release.jl
@@ -128,7 +128,7 @@ install -v -m 755 "uno_ampl${exeext}" -t "${bindir}"
 
 # Currently, Uno does not provide a shared library. This may be useful in the future once it has a C API.
 # We just check that we can generate it, but we don't include it in the tarballs.
-${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o libuno.${dlext} -L${libdir} -l${OMP} -lopenblas -ldmumps -lmetis -lhsl -lhighs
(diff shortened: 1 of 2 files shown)


cvanaret (approved):

cvanaret:
~~Looks like it's working :)~~ Spoke too fast.

amontoison:
@cvanaret 
I have an explanation of what happened.
I just checked the symbols in the static library `libbqpd.a`, and `alphac_` and `wsc_` are of type `"C"`, which likely means they are common blocks.
They are not marked as `"T"` like regular exported symbols.

If we use the C++ linker (with `g++`) instead of the Fortran linker (with `gfortran`), maybe these symbols are not included in `libuno.a` (specific to Fortran).
This would explain why we need to link explicitly with `-lbqpd` when creating the shared library.

```shell
alexis@HP-Spectre:~/Bureau/git/bqpd/builddir$ nm libbqpd.a | grep C
0000000000000018 C epsc_
0000000000000004 C iprintc_
0000000000000008 C minorc_
0000000000000004 C noutc_
0000000000000020 C alphac_
0000000000000034 C bqpdc_
0000000000000004 C iprintc_
0000000000000004 C noutc_
0000000000000008 C vstepc_
0000000000000018 C epsc_
0000000000000018 C factorc_
0000000000000004 C iprintc_
0000000000000004 C mxm1c_
0000000000000004 C noutc_
0000000000000008 C refactorc_
000000000000006c C sparsec_
0000000000000018 C wsc_
```

amontoison:
> ~Looks like it's working :)~ Spoke too fast.

I opened #200 to fix the issue. The linker is a little bit blind on Windows and needs some help to find the static libraries...

PR #205, opened 2025-07-03
Use the old linker on Mac

amontoison:
I got a similar issue with HSL.

### DIFF
--- a/.github/julia/build_tarballs_release.jl
+++ b/.github/julia/build_tarballs_release.jl
@@ -175,4 +175,5 @@ build_tarballs(
     dependencies;
     julia_compat = "1.6",
     preferred_gcc_version = v"10.2.0",
+    clang_use_lld=false,
 )
--- a/.github/julia/build_tarballs_yggdrasil.jl
+++ b/.github/julia/build_tarballs_yggdrasil.jl
@@ -116,4 +116,5 @@ build_tarballs(
     dependencies;
     julia_compat = "1.9",
     preferred_gcc_version = v"10.2.0",
+    clang_use_lld=false,
 )


cvanaret:
Valuable experience ;) I should definitely add a CI build script for Darwin so that we can catch this kind of errors earlier.

amontoison:
@cvanaret Do you hardcoded a link `-lgcc` in Uno?
On Mac the default C / C++ compiler is clang / clang++.

cvanaret:
Mh I had a look at the CMakeLists, it looks ok to me.

cvanaret:
Could it be because of
`[06:35:49] -- The Fortran compiler identification is GNU 10.2.0`
?

amontoison:
I will have a look this weekend.
I just arrived at Chicago. Long journey from Kiel today...
I also have a backlog of four weeks of emails to answer. 😮‍💨

We can have `-lgcc` on Mac but we need to install CompilerSupportLibraries_jll for that and we don't do it to not have conflicts between `ligomp` and `libomp`.
Because we use clang / clang++ on Mac, we also don't any flag like `-lgcc`, so we need to understand why CMake uses this flag.

For the Fortran compiler I expect that they add `-lgfortran` and not `-lgcc`.
I will investigate.

cvanaret:
Ah! Good luck with catching up 🫣 It's not very urgent tbh, so only when you have a moment!

[This line](https://github.com/cvanaret/Uno/blob/a284037b899ee641b973726e1782b2ee4d25783a/CMakeLists.txt#L225) might be the culprit. It adds extra flags (such as `-lquadmath`, see [here](https://cmake.org/pipermail/cmake/2017-July/065778.html)), so I'm wondering whether it may have added these GCC flags.

amontoison:
You are right Charlie, the culprit is the line
```shell
target_link_libraries(uno PUBLIC ${LIBRARIES} ${CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES})
```
I displayed the content in #206 and we can see `gcc`, which is not good on Mac where we mix `clang++`, `clang` and `gfortran`.
```shell
CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES = emutls_w;heapt_w;gfortran;gcc;quadmath
```

cvanaret:
Fixed in https://github.com/cvanaret/Uno/pull/206.

PR #206, opened 2025-07-05
Adapt CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES for Mac

amontoison:
https://github.com/amontoison/Uno/actions/runs/16089175439/job/45404079470

### DIFF
--- a/.github/julia/build_tarballs_release.jl
+++ b/.github/julia/build_tarballs_release.jl
@@ -175,5 +175,5 @@ build_tarballs(
     dependencies;
     julia_compat = "1.6",
     preferred_gcc_version = v"10.2.0",
-    clang_use_lld=false,
+    # clang_use_lld=false,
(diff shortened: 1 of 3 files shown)


cvanaret (approved):

cvanaret:
Thanks for investigating Alexis! This should also solve the problem with emutls that @leyffer had when he visited with a Mac ✌️

amontoison:
@cvanaret I only tested locally `aarch64-apple-darwin` with this PR.
I just checked and the old linker `ld` is still needed with `x86_64-apple-darwin`...
Apple broke the compilation on Mac-Intel with `lld`.

I will open a last PR and you can retag the release 2.0.0 after that.
I tested all platforms.
For the next release 2.x, I propose to add more CI tests related to the compilation :)

PR #208, opened 2025-07-05
Use ld instead of lld for Mac-Intel

amontoison:
Should be good for the release 2.0 after that :metal:


cvanaret:
It's a wrap! https://github.com/cvanaret/Uno/releases/tag/v2.0.0
Thanks a lot @amontoison, you rock! Beer's on me at ICCOPT!

Issue #35, opened 2024-10-19
Install Uno via CMake install


cvanaret:
Add support to CMake install.

cvanaret:
Hi @amontoison, do we need to install a static or a shared library?

amontoison:
It depends what is the CMake option used by the user (`-DBUILD_SHARED_LIBS=ON` or `-DBUILD_SHARED_LIBS=OFF`)
But you should be able to handle all cases with something like:
```
install(TARGETS ${_target}
  COMPONENT libuno
  EXPORT ${_target}Targets
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}  # DLLs on windows
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}  # Static libraries
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}  # Shared libraries
  PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```

amontoison:
@cvanaret I think that your current build system only generate static libraries.
CMake could do the shared for you but you probably need a few modifications.

cvanaret:
Thanks for the example.
If I wanted both static and shared libraries to have the same name (not sure it's portable though), I could use something like:
```cmake
add_library(uno_lib OBJECT ${UNO_SOURCE_FILES})
set_property(TARGET uno_lib PROPERTY POSITION_INDEPENDENT_CODE ON)

add_library(uno_shared SHARED $<TARGET_OBJECTS:uno_lib>)
set_target_properties(uno_shared PROPERTIES OUTPUT_NAME uno)

add_library(uno_static STATIC $<TARGET_OBJECTS:uno_lib>)
set_target_properties(uno_static PROPERTIES OUTPUT_NAME uno)
```

Do we want to compile with PIC for the shared library and without for the static library? It'll be expensive but it most likely results in the most efficient static library.

amontoison:
It's fine if they have the same name because the extension will be different.
You can't generate a shared library without PIC so we need it for that. In Meson, it compiles only the code once when we ask both libraries (shared and static) so you should probably need to hack something to force CMake to compile Uno twice with different compilation options.

PR #195, opened 2025-07-01
[WIP] Python bindings

cvanaret:

PR #213, opened 2025-07-08
[WIP] Compiling static and shared libraries of Uno + make install

cvanaret:
First tests with:
- compiling static and shared libraries of Uno
- `make install`

Made most targets `EXCLUDE_FROM_ALL`, which means that we need an explicit `make target` command to build them.

Fixes https://github.com/cvanaret/Uno/issues/35.
Shared libraries are necessary for the Python bindings (https://github.com/cvanaret/Uno/pull/195).


amontoison:
@cvanaret What is the library (shared or static) used for the "AMPL" binary?

Did you fixed the `make install`?
I can easily test this part with the cross-compiler.

cvanaret:
@amontoison I set the correct targets for the builds.
Now I need to fix the `install`. Clearly, `if (TARGET uno_***)` [here](https://github.com/cvanaret/Uno/pull/213/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR281) doesn't work:

```cmake
[ 98%] Built target uno_files
[100%] Built target uno_static
Install the project...
-- Install configuration: "Debug"
-- Up-to-date: /usr/local/lib/libuno.a
CMake Error at cmake_install.cmake:56 (file):
  file INSTALL cannot find "/media/data/Uno/build_debug/libuno.so": No such
  file or directory.
make: *** [Makefile:100: install] Error 1
```

cvanaret:
@amontoison I ended up using the flag `BUILD_SHARED_LIBS`.
Can you try `make install`?

amontoison:
@cvanaret I tested `make install` with the static library and the binary `uno_ampl` is not installed.

cvanaret:
Ah, I thought we'd only install `libuno`. Ok I'll add `uno_ampl`.

cvanaret:
I came up with [this solution](https://github.com/cvanaret/Uno/pull/213/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR279), but it doesn't run on my machine (when `uno_ampl` hasn't been built, it tries to install it anyway).

amontoison:
@cvanaret 
I suggest to have both options `BUILD_SHARED_LIBS` and `BUILD_STATIC_LIBS` and ensure that at list one of them is `"ON"`.
```shell
# Options
option(BUILD_STATIC_LIBS "Build using static libraries" ON)
option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
option(WITH_GTEST "Enable GoogleTest" OFF)

if(NOT BUILD_STATIC_LIBS AND NOT BUILD_SHARED_LIBS)
  message(FATAL_ERROR "At least one of BUILD_SHARED_LIBS or BUILD_STATIC_LIBS must be ON.")
endif()
```
To check if `uno_ampl` was created, you can do something like this:
```shell
if(BUILD_STATIC_LIBS)
  add_library(uno_static STATIC ${UNO_SOURCE_FILES})
  target_include_directories(uno_static PUBLIC ${DIRECTORIES})
  target_link_libraries(uno_static PUBLIC ${LIBRARIES} ${FORTRAN_LIBS})
  set_target_properties(uno_static PROPERTIES OUTPUT_NAME "uno" POSITION_INDEPENDENT_CODE ON)
endif()

if(BUILD_SHARED_LIBS)
  add_library(uno_shared SHARED ${UNO_SOURCE_FILES})
  target_include_directories(uno_shared PUBLIC ${DIRECTORIES})
  target_link_libraries(uno_shared PUBLIC ${LIBRARIES} ${FORTRAN_LIBS})
  set_target_properties(uno_shared PROPERTIES OUTPUT_NAME "uno")
endif()

...

set(INSTALL_TARGETS "")
if(TARGET uno_static)
   list(APPEND INSTALL_TARGETS uno_static)
endif()
if(TARGET uno_shared)
   list(APPEND INSTALL_TARGETS uno_shared)
endif()
if(TARGET uno_ampl)
   list(APPEND INSTALL_TARGETS uno_ampl)
endif()
```

cvanaret:
Got it! Is there a clear advantage over:
```shell
cmake -DBUILD_SHARED_LIBS=ON ..
make
cmake -DBUILD_SHARED_LIBS=OFF ..
make
```
?

amontoison:
> Got it! Is there a clear advantage over:
> 
> ```shell
> cmake -DBUILD_SHARED_LIBS=ON ..
> make
> cmake -DBUILD_SHARED_LIBS=OFF ..
> make
> ```
> 
> ?

Yes, you compile the object files (*.o) only once. You don't duplicate the work. It requires only a few seconds / minutes for Uno to compile (now). But for example, for GALAHAD it is not the same story with 2 millions of lines of Fortran.

You also have more control on what library of Uno is used for the AMPL binary.
The idea is to used the static library if possible instead of the shared one to have the binary standalone as much as possible.
But you can still use the shared library if you only have this one.

cvanaret:
OK, so I need to use the trick I described [here](https://github.com/cvanaret/Uno/issues/35#issuecomment-2537291747). This is the way to avoid duplicate compilation of the objects.
I'm on it!

amontoison:
@cvanaret 
A good test is to replace the following content in `build_tarballs.jl`:
```shell
make -j${nproc}
make uno_ampl -j${nproc}
# Uno does not support `make install`. Manually copy for now.
install -v -m 755 "uno_ampl${exeext}" -t "${bindir}"
# Currently, Uno does not provide a shared library. This may be useful in the future once it has a C API.
# We just check that we can generate it, but we don't include it in the tarballs.
${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o libuno.${dlext} -L${prefix}/lib -lbqpd -L${libdir} -l${OMP} -l${LBT} -ldmumps -lmetis -lhsl -lhighs -lgfortran
# cp libuno.${dlext} ${libdir}/libuno.${dlext}
```
by
```shell
make -j${nproc}
make install
```
Add active the compilation of both static and shared libraries.
At the end of the cross-compilation, we have an "audit" that is checking if everything is correctly linked / installed.

amontoison:
The linker on Linux is permissive but not on Mac and Windows :)
You don't link Uno correctly to HSL, HiGHS and MUMPS.

amontoison:
Also in `build_tarballs_yggdrasil.jl`, you should use the same products than the the one used for `build_tarballs_release.jl`.
```julia
   ExecutableProduct("uno_ampl", :amplexe),
   LibraryProduct("libuno", :libuno),
   FileProduct("lib/libuno.a", :libuno_a),
```
Even if we only use `uno_ampl` for the Julia tests, it helps us to check that the compilation / installation work as expected.

cvanaret:
I fixed the Ygg products and the HIGHS cache variable. Need to see the error messages for the other libraries.

cvanaret:
@amontoison is MUMPS' `libmpiseq` also in `${libdir}`? On my machine, it's in a `libseq` directory.

amontoison:
From what I remmeber yes. I only have a `libseq` folder for the header files. I can check in 30 minutes.
We just need to open a tarball of `MUMPS_seq_jll.jl` to check that.

amontoison:
@cvanaret 
I confirm that everything is in `${libdir}`.
Capture d’écran du 2025-07-09 19-27-48
Capture d’écran du 2025-07-09 19-27-35

cvanaret:
@amontoison looking good!

amontoison (approved):
LGTM! I am scared to try the compilation of the shared library with MSVC but I can be relevant to check that the build system is very robust.

cvanaret:
Ok, I'll try building the shared lib with MSVC, fingers crossed.
Well that failed fast.

amontoison:
I propose to merge with the current modifications.

imciner2:
So, the MSVC failures at link stage:
```
 LINK Pass 1: command "C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\uno_shared.dir\objects1.rsp /out:uno.dll /implib:uno.lib /pdb:D:\a\Uno\Uno\uno.pdb /dll /version:0.0 /machine:x64 /debug /INCREMENTAL deps\lib\libbqpd.a deps\lib\libdmumps.a deps\lib\libmumps_common.a deps\lib\libpord.a deps\lib\libmpiseq.a deps\lib\libmetis.a deps\lib\liblapack.a deps\lib\libblas.a kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:CMakeFiles\uno_shared.dir/intermediate.manifest CMakeFiles\uno_shared.dir/manifest.res" failed (exit code 1120) with the following output:
BQPDSolver.cpp.obj : error LNK2019: unresolved external symbol BQPD referenced in function "private: void __cdecl uno::BQPDSolver::solve_subproblem(class uno::Subproblem const &,class uno::Vector<double> const &,class uno::Direction &,class uno::WarmstartInformation const &)" (?solve_subproblem@BQPDSolver@uno@@AEAAXAEBVSubproblem@2@AEBV?$Vector@N@2@AEAVDirection@2@AEBVWarmstartInformation@2@@Z)
BQPDSolver.cpp.obj : error LNK2001: unresolved external symbol WSC
MUMPSSolver.cpp.obj : error LNK2019: unresolved external symbol dmumps_c referenced in function "public: __cdecl uno::MUMPSSolver::MUMPSSolver(void)" (??0MUMPSSolver@uno@@QEAA@XZ)
```
are probably because you are mixing the dependency libraries compiled with GCC (because you download the static libraries from a JLL, which is a mingw/GCC build) and your code compiled with MSVC.

When linking, the compilers have name mangling they perform to generate the actual symbol they store in the library (https://en.wikipedia.org/wiki/Name_mangling#), however the C++ standard doesn't actually specify how names should be mangled, and so there are compiler differences - most notably between GCC and MSVC. This means C++ code compiled with one is not visible to the linker of the other.

You will need to get the dependencies also built with MSVC in order to build an MSVC Uno version.

cvanaret:
Thanks for your input, Ian! We'll try to address this in a future PR.

imciner2:
Oh, another thing you might have to do for a shared library on Windows is to define the API explicitly. On Linux, the default is all symbols are visible in a shared library, but Windows generally is the other way, with all symbols hidden unless exported. This is done via an attribute on each class/function that gets exports. You can see in OSQP the logic to define that attribute is https://github.com/osqp/osqp/blob/master/include/public/osqp_export_define.h, and then we apply it to the functions https://github.com/osqp/osqp/blob/310d8cfbc2cb00df9ceb4829c78bfeac41f2cab7/include/public/osqp_api_functions.h#L40.

PR #218, opened 2025-07-11
Test shared libraries with MSVC

amontoison:
I also removed the dependencies because of the incompatibilities.


amontoison:
Good idea, I can do that.

PR #224, opened 2025-07-16
Julia tests for the filtersqp preset

cvanaret:
- [x] split the existing Julia tests for the `ipopt` and `filterslp` presets
- [ ] add Julia tests for the `filtersqp` preset

### DIFF
--- a/.github/julia/runtests_uno_filterslp.jl
+++ b/.github/julia/runtests_uno_filterslp.jl
@@ -1,0 +1,66 @@
+# Copyright (c) 2018-2025: Charlie Vanaret and contributors
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+# For help with this file, please contact `@odow` on GitHub.
+
+using Test
+
+import AmplNLWriter
+import MathOptInterface as MOI
+import MINLPTests
+import Uno_jll
+
+"""
+    Optimizer()
+
+Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
+solver.
+"""
+function Optimizer(options)
+    return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
+end
+
+Optimizer_Uno_filterslp() = Optimizer(["logger=SILENT", "preset=filterslp", "LP_solver=HiGHS", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
+
+# This testset runs https://github.com/jump-dev/MINLPTests.jl
+@testset "MINLPTests" begin
+    primal_target = Dict(
+        MINLPTests.FEASIBLE_PROBLEM => MOI.FEASIBLE_POINT,
+        # If Uno starts writing a .sol file with an infeasible point, change
+        # this to `=> MOI.INFEASIBLE_POINT`
+        MINLPTests.INFEASIBLE_PROBLEM => MOI.NO_SOLUTION,
+    )
+    # This function tests (potentially) non-convex nonlinear programs. The tests
+    # are meant to be "easy" in the sense that most NLP solvers can find the
+    # same global minimum, but a test failure can sometimes be allowed.
+    MINLPTests.test_nlp_expr(
+        Optimizer_Uno_filterslp;
+        exclude = [
+            "001_010",  # Local solution
+            "003_014",  # Local solution
+            "008_010",  # Local solution
+            # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
+            "005_010",
+            # Okay to exclude forever: AmplNLWriter does not support
+            # user-defined functions.
+            "006_010",
+            # Remove once https://github.com/cvanaret/Uno/issues/38 is fixed
+            "007_010",
+        ],
+        primal_target,
+        objective_tol = 1e-4,
+        primal_tol = 1e-4,
+    )
+    # This function tests convex nonlinear programs. Test failures here should
+    # never be allowed, because even local NLP solvers should find the global
+    # optimum.
+    MINLPTests.test_nlp_cvx_expr(
+        Optimizer_Uno_filterslp; 
+        primal_target,
+        objective_tol = 1e-4,
+        primal_tol = 1e-4,
+        exclude = ["501_011"],  # Iteration limit
+    )
+end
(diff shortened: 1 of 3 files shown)


cvanaret:
Hi @odow @amontoison,
I get a weird Julia test failure [here](https://github.com/cvanaret/Uno/actions/runs/16319262101/job/46092591296?pr=224) of Uno-filtersqp on the instance [nlp-expr/007_010](https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/007_010.jl).  The instance is infeasible, is detected so by Uno, and the AMPL solution reads:
```ampl
Uno 2.0.0: Infeasible stationary point

Options
3
0
1
0
2
2
2
2
0
0
1.526378828812082e-09
0.5
objno 0 200
suffix 4 0 18 0 0
upper_bound_duals
suffix 4 0 18 0 0
lower_bound_duals
```

200 is the error code for infeasible models, see top of p283 in the [AMPL manual](https://ampl.com/wp-content/uploads/BOOK.pdf).
When I call `solution_summary(model)`, I get:
```julia
solution_summary(; result = 1, verbose = false)
├ solver_name          : AmplNLWriter
├ Termination
│ ├ termination_status : LOCALLY_INFEASIBLE
│ ├ result_count       : 1
│ └ raw_status         : Uno 2.0.0: Infeasible stationary point
├ Solution (result = 1)
│ ├ primal_status        : UNKNOWN_RESULT_STATUS
│ ├ dual_status          : NO_SOLUTION
│ ├ objective_value      : 0.00000e+00
│ └ dual_objective_value : 0.00000e+00
└ Work counters
  └ solve_time (sec)   : 2.42654e-01
```
so JuMP has no information related to the primal solution (the `1.526378828812082e-09 0.5` in the AMPL solution, which corresponds to a local minimizer of the constraint violation) and the test fails here:
```julia
_check_status_subset(JuMP.primal_status(model), primal_target[problem_type])
```

How is that possible, given that the AMPL solution contains all the information? Is it the wrong kind of test [here](https://github.com/jump-dev/MINLPTests.jl/blob/master/src/nlp-expr/007_010.jl#L31)?

.github/julia/runtests_uno_filtersqp.jl
```diff
@@ -0,0 +1,120 @@
...
+    return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
+end
+
+Optimizer_Uno_filtersqp() = Optimizer(["logger=SILENT", "preset=filtersqp", "QP_solver=BQPD", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
+
+# This testset runs https://github.com/jump-dev/MINLPTests.jl
+@testset "MINLPTests" begin
+    primal_target = Dict(
+        MINLPTests.FEASIBLE_PROBLEM => MOI.FEASIBLE_POINT,
+        MINLPTests.INFEASIBLE_PROBLEM => MOI.NO_SOLUTION,
```
odow: @cvanaret the issue is here. JuMP expects that if Uno has found an infeasible problem, it will not return a primal solution.

(See the comment in your filterslp above)

.github/julia/runtests_uno_filtersqp.jl
```diff
@@ -0,0 +1,120 @@
...
+        exclude = [
+            "003_014",  # Local solution
+            "004_010",  # Local solution
+            "004_011",  # Local solution
+            "008_010",  # Local solution
+            # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
+            "005_010",
+            # Okay to exclude forever: AmplNLWriter does not support
+            # user-defined functions.
+            "006_010",
```
odow, line 48:
```suggestion
            "006_010",
            "007_010",
```

odow: Just skip the test

.github/julia/runtests_uno_filtersqp.jl
cvanaret: Got it. I saw the comment and tried both `MOI.NO_SOLUTION` and `MOI.INFEASIBLE_POINT` but it made no difference. I'll just skip the test. Thanks!

odow: It didn't make a difference because the value was UNKNOWN_RESULT_STATUS

.github/julia/runtests_uno_filtersqp.jl
```diff
@@ -0,0 +1,124 @@
...
+            r"^test_conic_NormInfinityCone_INFEASIBLE$",
+            r"^test_conic_NormOneCone_INFEASIBLE$",
+            r"^test_conic_linear_INFEASIBLE$",
+            r"^test_conic_linear_INFEASIBLE_2$",
+            r"^test_linear_INFEASIBLE$",
+            r"^test_linear_INFEASIBLE_2$",
+            r"^test_quadratic_SecondOrderCone_basic$",
+            r"^test_quadratic_nonconvex_constraint_basic$",
+            r"^test_solve_DualStatus_INFEASIBILITY_CERTIFICATE_",
+            r"^test_solve_VariableIndex_ConstraintDual_MAX_SENSE$",
```
odow, line 104: Why was this one skipped? It is a very suspicious failure

cvanaret: `test_quadratic_SecondOrderCone_basic` and `test_quadratic_nonconvex_constraint_basic` are reported as locally infeasible. The latter is nonconvex so it makes sense. For the former, starting at $(0, 0, 0)$, the QP at iteration 2 is infeasible and the current point $(0.5, 0.5, 0)$ turns out to be a minimum of constraint violation.
If we started from a feasible point, I think that by convexity all the QPs would be feasible.

odow: No I meant `test_solve_VariableIndex_ConstraintDual_MAX_SENSE`. It is not complicated:

https://github.com/jump-dev/MathOptInterface.jl/blob/d66c13dab9cbd76c3451faffafd7d828a4e04ee1/src/Test/test_solve.jl#L292-L311

cvanaret: Let me try again. Perhaps I'm writing the AMPL solution with incorrect signs.
https://github.com/cvanaret/Uno/pull/228

PR #228, opened 2025-07-17
Julia CI for filtersqp: included an MOI test that shouldn't fail

cvanaret:
See https://github.com/cvanaret/Uno/pull/224#discussion_r2214398702.

### DIFF
--- a/.github/julia/runtests_uno_filtersqp.jl
+++ b/.github/julia/runtests_uno_filtersqp.jl
@@ -101,7 +101,6 @@ end
             r"^test_quadratic_SecondOrderCone_basic$",
             r"^test_quadratic_nonconvex_constraint_basic$",
             r"^test_solve_DualStatus_INFEASIBILITY_CERTIFICATE_",
-            r"^test_solve_VariableIndex_ConstraintDual_MAX_SENSE$",
             # ==================================================================
             # The following tests are okay to exclude forever.
             #
--- a/bindings/AMPL/AMPLModel.cpp
+++ b/bindings/AMPL/AMPLModel.cpp
@@ -309,6 +309,12 @@ namespace uno {
          std::string message = "Uno ";
          message.append(Uno::current_version()).append(": ").append(iterate_status_to_message(iterate_status));
          write_sol_ASL(this->asl, message.data(), iterate.primals.data(), iterate.multipliers.constraints.data(), &option_info);
+
+         // flip the signs of the multipliers and the objective back if we maximize
+         iterate.multipliers.constraints *= this->objective_sign;
+         iterate.multipliers.lower_bounds *= this->objective_sign;
+         iterate.multipliers.upper_bounds *= this->objective_sign;
+         iterate.evaluations.objective *= this->objective_sign;
       }
    }
 
--- a/uno/ingredients/subproblem_solvers/BQPD/BQPDSolver.cpp
+++ b/uno/ingredients/subproblem_solvers/BQPD/BQPDSolver.cpp
@@ -313,13 +313,24 @@ namespace uno {
       for (size_t active_constraint_index: Range(number_variables - static_cast<size_t>(this->k))) {
          const size_t index = static_cast<size_t>(std::abs(this->active_set[active_constraint_index]) - this->fortran_shift);
 
+         // bound constraint
          if (index < number_variables) {
-            // bound constraint
-            if (0 <= this->active_set[active_constraint_index]) { // lower bound active
-               direction_multipliers.lower_bounds[index] = this->residuals[index];
+            // variable is not fixed
+            if (this->lower_bounds[index] < this->upper_bounds[index]) {
+               if (0 <= this->active_set[active_constraint_index]) { // lower bound active
+                  direction_multipliers.lower_bounds[index] = this->residuals[index];
+               }
+               else { // upper bound active
+                  direction_multipliers.upper_bounds[index] = -this->residuals[index];
+               }
             }
-            else { // upper bound active */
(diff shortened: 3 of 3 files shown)


odow:
> Perhaps I'm writing the AMPL solution with incorrect signs

Looks like that :smile:

Yay for tests

cvanaret:
Now, figure out why 3000 tests pass and a single one fails :smiley: 

What's weird is that for the JuMP model, the duals are correct:

```julia
using JuMP, AmplNLWriter, Uno_jll

function Optimizer()
   options = String["preset=filtersqp", "print_solution=yes"]
end
       
model = Model(() -> Optimizer());

@variable(model, x);
@objective(model, Max, x);
xl = @constraint(model, x >= 1);
xu = @constraint(model, x <= 1);
optimize!(model);

atol = 0.0001 # MOI tolerance
# get duals
sl = dual(xl)
su = dual(xu)
# test optimality conditions
sl + su == -1
sl >= -atol
su <= atol
```

For some reason, MOI's `sl` is $-1$ (while JuMP's is $0$ :open_mouth: ), which fails the test `sl >= -atol`.
Either I'm missing something, or there might be a bug in MOI, right?

odow:
You're missing something :smile:. The duals tested are on variable bounds, not affine constraints added with `@constraint`.

```Julia
julia> using JuMP, AmplNLWriter, Uno_jll

julia> function Optimizer()
          options = String["preset=filtersqp", "print_solution=yes"]
       end
Optimizer (generic function with 1 method)

julia> model = Model(Optimizer)
A JuMP Model
├ solver: AmplNLWriter
├ objective_sense: FEASIBILITY_SENSE
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> @variable(model, 1 <= x <= 1)
x

julia> @objective(model, Max, x)
x

julia> optimize!(model)
Variable x0 has identical bounds
Original model /var/folders/bg/dzq_hhvx1dxgy6gb5510pxj80000gn/T/jl_lpavN5/model.nl
1 variables, 0 constraints (0 equality, 0 inequality)
Reformulated model /var/folders/bg/dzq_hhvx1dxgy6gb5510pxj80000gn/T/jl_lpavN5/model.nl
1 variables, 0 constraints (0 equality, 0 inequality)

Used overwritten options:
- LP_solver = BQPD
- TR_min_radius = 1e-8
- TR_radius = 10
- globalization_mechanism = TR
- hessian_model = exact
- inequality_handling_method = inequality_constrained
- loose_tolerance = 1e-6
- print_solution = yes
- progress_norm = L1
- protect_actual_reduction_against_roundoff = no
- regularization_strategy = none
- residual_norm = L2
- tolerance = 1e-6

─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 iter   TR iter  penalty     TR radius    step norm   objective   stationarity  complementarity  status          
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 0      -        -           1.0000e+01   -           -1.0000e+00 1.0000e+00    0.0000e+00       initial point   
 1      1        -           1.0000e+01   0.0000e+00  -1.0000e+00 0.0000e+00    0.0000e+00       0 primal step   
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 iter   TR iter  penalty     TR radius    step norm   objective   stationarity  complementarity  status          
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Uno 2.0.0 (TR l1-merit inequality-constrained method with exact Hessian and no regularization)
Sat Jul 19 08:10:24 2025
────────────────────────────────────────
Optimization status:			Success
Iterate status:				Feasible KKT point
Objective value:			1
Primal feasibility:			0
┌ Stationarity residual:		0
└ Complementarity residual:		0
┌ Feasibility stationarity residual:	0
└ Feasibility complementarity residual:	0
┌ Infeasibility measure:		0
│ Objective measure:			-1
└ Auxiliary measure:			0
Primal solution:			1 
┌ Constraint multipliers:		
│ Lower bound multipliers:		1 
└ Upper bound multipliers:		-0 
┌ Constraint feasibility multipliers:	
│ Lower bound feasibility multipliers:	0 
└ Upper bound feasibility multipliers:	0 
Objective multiplier:			1
CPU time:				0.010948s
Iterations:				1
Objective evaluations:			2
Constraints evaluations:		0
Objective gradient evaluations:		2
Jacobian evaluations:			0
Hessian evaluations:			1
Number of subproblems solved:		1

julia> atol = 0.0001 # MOI tolerance
0.0001

julia> sl = dual(LowerBoundRef(x))
-1.0

julia> su = dual(UpperBoundRef(x))
-0.0

julia> sl + su == -1
true

julia> sl >= -atol
false

julia> su <= atol
true
```

The dual should be on the upper bound, not the lower bound.

cvanaret:
Ah, me wrongly interpreting the example makes more sense :yum: 
Thanks! It actually looks like a bug in BQPD!
@leyffer :wave:

odow:
For why it's one test in 3000: I assume it's the only test with Max and a fixed variable and we explicitly test the dual of both bounds. Getting it wrong is an easy mistake to make.

cvanaret:
Absolutely and the AMPL interface automatically eliminates fixed variables, so the bug may go unnoticed.

odow:
All fixed? So not a bug in BQPD?

cvanaret:
It looks like the duals in BQPD must be interpreted another way when the bounds are identical. So not a bug - as always: user error 😀

odow:
Ha. This is why MOI has such a strong dual convention and tests to enforce it. There are soooooooo many variations between and within solvers.

PR #229, opened 2025-07-23
Dense objective gradient

cvanaret:
Until now, the objective gradient was sparse, which was useless. It is now dense.
This will simplify the C interface.

### DIFF
--- a/.github/julia/runtests_uno_ipopt.jl
+++ b/.github/julia/runtests_uno_ipopt.jl
@@ -22,7 +22,7 @@ function Optimizer(options)
     return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
 end
 
-Optimizer_Uno_ipopt() = Optimizer(["logger=INFO", "preset=ipopt", "unbounded_objective_threshold=-1e15", "print_solution=yes"])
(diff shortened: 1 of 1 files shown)


amontoison (approved):

mjacobse:
I can't really think of a good use of the sparsity information of the objective gradient in the solver itself either, but at least for approximations of the derivatives I believe there are some use cases. For instance when using finite differences to compute the objective gradient one can avoid function evaluations for components that are known to be zero. Or we can also get an upper bound on the sparsity of the objective Hessian from just the sparsity of the objective gradient.

Figured I could share the thought just in case. Sorry to bother you if you already considered and disregarded this as something not to be implemented in Uno itself. I could indeed imagine that this might work better with such approximations being done externally, separate from the solver. And even if there is some potential future use of the sparsity of the objective gradient in the solver itself, I suppose that should also use the separation mentioned in #231 .

cvanaret:
Hi @mjacobse,
You're right, if I were to write a derivative checker with finite differences, it would be beneficial to have the sparsity pattern of the objective gradient.
@amontoison and I decided on a basic API to define models that doesn't include the objective gradient sparsity at the moment, but it would be easy to add it at some future point.

Issue #231, opened 2025-07-23
Separate the sparsity and the values evaluation in Jacobian and Hessian calls


cvanaret:
Separate the Jacobian and Hessian calls into:
- sparsity computation: `compute_jacobian_sparsity(row_indices, column_indices)`. Will be called once at the beginning.
- evaluation of numerical values: `evaluate_jacobian_values(x, jacobian)`.
- same for Hessian.

cc @amontoison

PR #232, opened 2025-07-23
Compute the Jacobian and Hessian sparsity patterns once and for all

cvanaret:
- [x] compute the Hessian structure
- [x] call it once and for all the beginning
- [x] compute the constraint Jacobian structure
- [x] call it once and for all the beginning
- [x] have the subproblem solvers report whether they expect the constraint Jacobian as row-major or column-major order
- [x] fix the computation of linearized constraints
- [x] convert COO format to CSR format for BQPD
- [x] convert COO format to CSC format for HiGHS
- [x] store the constraint Jacobian only in the subproblem solver
- [x] fix computation of Lagrangian gradient

Preprocessing was obsolete and was deleted. It will be rebuilt from scratch at a later stage.

Fixes https://github.com/cvanaret/Uno/issues/231


cvanaret:
Good news: the `filtersqp` preset with BQPD now runs.
I had to:
- sort the Jacobian entries so that the constraints are in increasing order (I compute a permutation vector once and for all).
- create the CSR sparsity from the COO sparsity.
- copy the numerical data with permutation from the COO format to the CSR format.

I think the HiGHS interface won't be too much work. IIRC, we can pass the Jacobian in column-major format, which is the preferred format.

cvanaret:
Fixed the initial computation of primal-dual residuals (the constraint Jacobian wasn't evaluated).

PR #237, opened 2025-07-28
Lagrangian sign convention in ASL: call the scaling function once instead of flipping the signs

cvanaret:
Uno uses the convention $\nabla^2 L(x, y) = \nabla^2 f(x) - \sum_j y_j \nabla^2 c_j(x)$, while ASL uses a + sign.
We now call `lagscale_ASL` once instead of flipping the signs of $y$ at every Hessian evaluation.


cvanaret:
@amontoison have you ever seen a CI error like that before?

amontoison:
@cvanaret GitHub is in maintenance, some downloads are failing because of that.
You can merge, it's safe.

cvanaret:
Ah I see! Thank you.

PR #242, opened 2025-07-29
[CI] Simplify scripts

amontoison:


amontoison:
I need to upload `Uno_jll.jl` as an artifact to make it work because each job is isolated. Let's keep things like this for now.

PR #254, opened 2025-08-05
Fix bad argv access with omitted -AMPL flag

mjacobse:
Seems to have been missed in b5a79a4ca. Got me the cryptic exception message "basic_string::_M_construct null not valid" when trying to run it without the `-AMPL` flag (I suppose `argv[2]` happens to be `nullptr` if `argc <= 2`).


cvanaret (approved):

cvanaret:
Of course, a rookie mistake! Thanks @mjacobse!

PR #256, opened 2025-08-05
Implement retry-loop for enlarging MA57 workspace

mjacobse:
React to failure of MA57 numeric factorization due to insufficient workspace size by creating new, larger workspace and copying data into the new arrays to continue factorization from where left off, as described in MA57 documentation.

Without this, I think what happens (at least for the ipopt preset) is that due to the numeric factorization failing, MA57 reports a rank of `0`. That causes the regularization loop to start increasing the regularization until MA57 needs to do so little pivoting that the factorization fits into the initial workspace size. So the regularization was kind of hiding this.

Noticed this with cvxqp1, which surprisingly triggered a lot of regularization and could not be solved within the 2000 iteration limit. With this patch it's solved in 38 iterations.


cvanaret:
Wow, that's a great catch! I'm curious to see how many instances turn out to be solvable now. Thanks a lot!

PR #263, opened 2025-08-10
[CI] Test the filtersqp preset + identity Hessian model

cvanaret:
Test Uno with the following options:
`preset=filtersqp hessian_model=identity QP_solver=BQPD`

### DIFF
--- a/.github/julia/runtests_uno_filtersqp_bqpd_identity.jl
+++ b/.github/julia/runtests_uno_filtersqp_bqpd_identity.jl
@@ -40,6 +40,8 @@ Optimizer_Uno_filtersqp() = Optimizer(["logger=SILENT", "preset=filtersqp", "hes
             "003_014",  # Local solution
             "004_010",  # Local solution
             "004_011",  # Local solution
+            # Remove once https://github.com/cvanaret/Uno/issues/39 is fixed
+            "005_010",
             "007_010",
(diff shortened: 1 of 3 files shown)


cvanaret:
The script for testing the `filtersqp` preset with identity Hessian (so essentially some kind of constrained gradient descent) now runs. However, the macOS build scripts started failing out of the blue. Perhaps @amontoison has an idea why?

amontoison:
You have the same error than what we have with Yggdrasil on Mac.
The culprit is this file:
```shell
/Users/runner/work/Uno/Uno/build/CMakeFiles/FortranCInterface/VerifyCXX
```
We do that in the cross-compiler:
```shell
# FortranCInterface_VERIFY fails on macOS, but it's not actually needed for the current build
sed -i 's/FortranCInterface_VERIFY(CXX)/# FortranCInterface_VERIFY(CXX)/g' ../CMakeLists.txt
```
I am wondering why you keep this file, it seems to be a can of worms.

cvanaret:
Ah! I see, thanks. I'm fine with dropping this line altogether, I'll open a PR.
https://github.com/cvanaret/Uno/pull/264

cvanaret:
It seems to be an issue with the migration to `macos-15`. It runs on `macos-14` (https://github.com/cvanaret/Uno/pull/264).

PR #268, opened 2025-08-11
[CI] Run filtersqp preset with QP solver HiGHS and primal regularization

cvanaret:
Added full set of MINLP.jl and MOI instances.


cvanaret:
HiGHS reads the diagonal of the Hessian to fail early in case of indefiniteness. This means we have to properly compress the entries in Uno when we use regularization.
Momentarily on hold.

Issue #230, opened 2025-07-23
C interface


cvanaret:
Required function signatures were discussed today with @amontoison.

PR #269, opened 2025-08-11
[WIP] C and Julia interfaces

cvanaret:
C interface:
- [x] write the interface `Uno_C_API.h`. All the functions will take as parameters the sizes of the vectors and an object `user_data` of type `void*`, and 
  - [x] objective
  - [x] constraints
  - [x] dense objective gradient
  - [x] constraint Jacobian
  - [x] Hessian
  - [x] Jacobian-vector product (with a boolean to indicate whether the `x` vector is new)
  - [x] Jacobian^T-vector product (with a boolean to indicate whether the `x` vector is new)
  - [x] Hessian-vector product (with a boolean to indicate whether the `x` vector is new)
  - [x] function to get the current Uno version
- [x] write the implementation `Uno_C_API.cpp`
  - [x] write classes `UserModel` (contains the description of the user model) and `UnoModel` (complies with the `uno::Model` interface)
    - [x] function pointers
    - [x] number of constraint Jacobian nonzeros
    - [x] number of Hessian nonzeros
    - [x] variables bounds
    - [x] constraints bounds
    - [x] indexing (C/Fortran)
    - [x] problem type (linear/quadratic/nonlinear)
    - [x] constraint Jacobian sparsity pattern
    - [x] Hessian sparsity pattern
    - [x] Jacobian operator (Jacobian-vector product)
    - [x] Jacobian transposed operator (Jacobian^T-vector product)
    - [x] Hessian-vector product
  - [x] write function `uno_create_model` that builds a `CModel` with partial information
    - [x] set objective + optimization sense
    - [x] set constraints
    - [x] set Hessian (required: Lagrangian convention, uplo)
    - [x] set user data
    - [x] set Jacobian operators
    - [x] set Hessian operator
    - [x] set initial primal-dual iterate
  - [x] write function `uno_create_default_options`, setters `uno_set_option`, and `uno_destroy_options`
  - [x] write function `uno_optimize` that calls the C++ library
  - [x] write getters to access the content of the optimization result
  - [x] write function `uno_destroy_model`
  - [x] handle error cases
- [x] CMake
  - [x] build `Uno_C_API.cpp` with Uno libraries
  - [x] install `Uno_C_API.h`
- [x] example file `example_hs015.c`

In Uno:
- [x] handle indexings
- [x] handle Lagrangian sign convention
- [x] handle min/maximization
- [x] handle the Hessian triangular part

Closes https://github.com/cvanaret/Uno/issues/230.
cc @amontoison @rujialiu


bindings/C/Uno.h
```diff
@@ -0,0 +1,59 @@
...
+
+   // constraint_jacobian_sparsity
+   // - takes as inputs two vectors "row_indices" and "column_indices", allocates memory for them, and stores the row and
+   // column indices of the constraint Jacobian entries in coordinate (COO) format.
+   void constraint_jacobian_sparsity(int* row_indices, int* column_indices, void* user_data);
+
+   // lagrangian_hessian_sparsity
+   // - takes as inputs two vectors "row_indices" and "column_indices", allocates memory for them, and stores the row and
+   // column indices of the Lagrangian Hessian entries in coordinate (COO) format.
+   void lagrangian_hessian_sparsity(int* row_indices, int* column_indices, void* user_data);
```
amontoison: How to provide the length of these arrays?

bindings/C/Uno.h
```diff
@@ -0,0 +1,59 @@
...
+   // objective_gradient
+   // - takes as inputs a vector "x" of size "number_variables" and an object "user_data", and stores the dense objective
+   // gradient at "x" in the vector "gradient" of size "number_variables".
+   // - returns an integer that is 0 if the evaluations succeeded, and positive otherwise.
+   int objective_gradient(const double* x, int number_variables, double* gradient, void* user_data);
+
+   // constraint_jacobian_sparsity
+   // - takes as inputs two vectors "row_indices" and "column_indices", allocates memory for them, and stores the row and
+   // column indices of the constraint Jacobian entries in coordinate (COO) format.
+   void constraint_jacobian_sparsity(int* row_indices, int* column_indices, void* user_data);
```
amontoison: How to provide the length of these arrays?

bindings/C/Uno.h
```diff
@@ -0,0 +1,59 @@
...
+   // lagrangian_hessian_sparsity
+   // - takes as inputs two vectors "row_indices" and "column_indices", allocates memory for them, and stores the row and
+   // column indices of the Lagrangian Hessian entries in coordinate (COO) format.
+   void lagrangian_hessian_sparsity(int* row_indices, int* column_indices, void* user_data);
+
+   // constraint_jacobian
+   // - takes as inputs a vector "x" of size "number_variables" and an object "user_data", and stores the entries of the
+   // sparse constraint Jacobian in the vector "jacobian" of size "number_nonzeros". The values should be in the same
+   // order as the indices provided in "constraint_jacobian_sparsity".
+   int constraint_jacobian(const double* x, int number_variables, double* jacobian, int number_nonzeros,
```
amontoison:
```suggestion
   int constraint_jacobian(const double* x, int number_variables, double* jacobian, int number_jacobian_nonzeros,
```

bindings/C/Uno.h
```diff
@@ -0,0 +1,59 @@
...
+      void* user_data);
+
+   // lagrangian_hessian
+   // - takes as inputs a vector "x" of size "number_variables", an objective multiplier, a vector "multipliers" of
+   // Lagrange multipliers of size "number_constraints", and an object "user_data", and stores the entries of the
+   // sparse Lagrangian Hessian in the vector "hessian" of size "number_nonzeros". The values should be in the same
+   // order as the indices provided in "constraint_jacobian_sparsity".
+   // Only the lower triangular part of the symmetric Lagrangian Hessian should be provided.
+   int lagrangian_hessian(const double* x, int number_variables, double objective_multiplier, const double* multipliers,
+      int number_constraints, double* hessian, int number_nonzeros, void* user_data);
```
amontoison:
```suggestion
      int number_constraints, double* hessian, int number_hessian_nonzeros, void* user_data);
```

bindings/C/Uno.h
```diff
@@ -0,0 +1,87 @@
...
+   int constraint_jacobian_vector_product(const double* x, int number_variables, bool evaluate_at_x,
+      const double* vector, double* result, int number_constraints, void* user_data);
+
+   // constraint_jacobian_transposed_vector_product
+   // - takes as inputs a vector "x" of size "number_variables", a boolean "evaluate_at_x" that indicates whether
+   // the Jacobian should be evaluated at "x" (otherwise, the current constraint Jacobian is used), a vector "vector"
+   // of size "number_constraints" and an object "user_data", and stores the Jacobian transposed-vector product in the
+   // vector "result" of size "number_variables".
+   // - returns an integer that is 0 if the evaluation succeeded, and positive otherwise.
+   int constraint_jacobian_transposed_vector_product(const double* x, int number_variables, bool evaluate_at_x,
```
amontoison: `number_constraints` is after `result` in `constraint_jacobian_vector_product` but before in `constraint_jacobian_transposed_vector_product`.

bindings/C/Uno.h
```diff
@@ -0,0 +1,87 @@
...
+      const double* vector, int number_constraints, double* result, void* user_data);
+
+   // lagrangian_hessian_vector_product
+   // - takes as inputs a vector "x" of size "number_variables", an objective multiplier, a vector "multipliers" of
+   // Lagrange multipliers of size "number_constraints", and an object "user_data", and stores the Lagrangian
+   // Hessian-vector product in the vector "result" of size "number_variables".
+   // - returns an integer that is 0 if the evaluation succeeded, and positive otherwise.
+   int lagrangian_hessian_vector_product(const double* x, int number_variables, double objective_multiplier,
```
amontoison: @cvanaret You forgot an input `"vector"` that represents the "vector" in lagrangian_hessian_vector_product ;)

amontoison: Should we also add a boolean `bool evaluate_at_x`?

bindings/C/Uno.h
cvanaret: Yes, a size follows the first vector that has this size. `result` has size `number_constraints` in `constraint_jacobian_vector_product`, while in `constraint_jacobian_transposed_vector_product`, `vector` has size `number_constraints`. Just a convention, but I could also e.g. pass both sizes as first two arguments.

bindings/C/Uno.h
cvanaret: IIRC, the modeler is in charge of allocating the vectors for the row and column indices? In which case, they should know how long the vectors should be. I will then copy these in my own (possibly longer) vectors.

bindings/C/Uno.h
cvanaret: Ahah yeah. Yes, I'll add a boolean :+1:

bindings/C/Uno.h
amontoison: Ok, but from a pointer you don't know always knows the length of the underlying array:
https://github.com/ralna/GALAHAD/blob/master/GALAHAD.jl/test/test_expo.jl#L38-L64
We can use the `user_data`for that but it is a little bit hacky.

bindings/C/Uno.h
amontoison: I will suggest to pass `int number_variables` (and `int number_constraints` where needed) as first arguments of all routines of the C interface. It is a simple convention easy to remember.

cvanaret: Works for me!

bindings/C/Uno.h
cvanaret: Mh `constraint_jacobian_sparsity` would create these vectors: https://github.com/ralna/GALAHAD/blob/master/GALAHAD.jl/test/test_expo.jl#L147
So either:
- we use `user_data` to get the Jacobian NNZ, or
- we pass the `row/column_indices` vectors directly to the model creation (not through a function call like I have in this PR), or
- we add a parameter NNZ to `constraint_jacobian_sparsity`, and call the function in the `CModel` with that argument (because the model is created with the Jacobian and Hessian NNZs)

What do you think?

amontoison: I like the third proposition. It is great to always provide the length of the arrays when we have them as argument;

cvanaret:
@amontoison should I turn my C signatures into typedefs, like they did in [IPOPT](https://github.com/coin-or/Ipopt/blob/8b54331f124d1e143a48e5c830ae63d605328d04/src/Interfaces/IpStdCInterface.h#L73)?

amontoison:
Yes, I will also recommend that. It is what I did a few days ago in GALAHAD:
https://github.com/ralna/GALAHAD/commit/b9239a2d5e8610ae00142a3e370f70020162b97a

amontoison:
@cvanaret For the symbols exported by the C interface, I suggest that we always use the prefix `uno_`, it is easy to remember and ensure that the symbols don't collapse with any symbol of other library.

What we didn't discussed to much is the API for the getter and setter. Should we pass everything as a pointer of void and recover it as a void **?

amontoison:
We should also not forget to add a function to get the version of Uno with three integers as input.

I complained to add it in libHSL, GALAHAD, Ipopt and SPRAL. It is very useful in high-level interfaces because we can easily check at runtime if the version is recent enough to support a given option and adapt our interface to support multiple releases of Uno if we do breaking changes.

Foe example, it is what we do with CUDA.jl.

-> https://github.com/coin-or/Ipopt/issues/824

rujialiu:
> @cvanaret For the symbols exported by the C interface, I suggest that we always use the prefix `uno_`, it is easy to remember and ensure that the symbols don't collapse with any symbol of other library.

Yes, avoiding name clashing is very important. I always had a hard time when trying to incoporate some old-style C libraries/C interfaces because of name clashes (macro, variable and functions) and global variable (including static variables).

> What we didn't discussed to much is the API for the getter and setter. Should we pass everything as a pointer of void and recover it as a void **?

Depending on the library, someitmes it's very handy to define some "bridge types" like this:

```c++
struct CommonParameters {
  int important_var1;
  double important_var2;
  ...
};
```

and offer some "combined" setter/getters. These bridge types are visible to the end users and considered part of the API.

It's benefical when
- there are lots of parameters and/or 
- they're logically groupped and/or 
- you can only validate the parameters when you finish setting the whole group
 
Last point: suppose if you need a.size() == b.size() == c.size(), you need to call this check 3 times, in `set_a()` and `set_b()` and `set_c()`. What's worse, in `set_a()` you might be unable/cumbersome to check whether user has already set b and c. Of course you can delay all checks just before you do actual work, but some people prefer to know validation error as early as possible.

The biggest drawback I can think of, is: it's harder to be backward-compatible (both API and ABI). But if we can get exactly uno version, it's not a big deal?

Of course, we can still avoid any `struct` and have one getter/setter for each thing. After all, this is C interface, typically people won't criticize is verbosity and we can always write wrappers on top of this. And I'm not an expert. So all of these above is just for reference only 😄

cvanaret:
@amontoison @rujialiu thanks for your inputs!

Here's the current state:
- I rewrote my function signatures as `typedef`s;
- I prefixed the function names with `uno_` (should I prefix the `typedef` names as well?);
- I added a function to get the current Uno version;
- calling `uno_create_model` creates an "empty" model with only the variables information, with no objective (= 0), no constraints, no Hessian, no user data. You can use setter functions to add an objective (+ gradient info), constraints (+ Jacobian info), a Lagrangian Hessian, or some user data. I took @rujialiu's suggestion to group stuff together (but without `struct`s).

Let me know how that looks.
I need to write `minimize` and `maximize` functions, then getters to access the content of the optimization result.

rujialiu:
Grouping stuffs without struct looks good! Should we prefix typedefs as well? Looking closer at Ipopt's code, I think they're not doing well, either. For example, `UserDataPtr` and `Eval_G_CB` are still prone to name clash (at least in theory 😆 ). Some people prefer to add a suffix like `_CB` (as did in Ipopt) or `_FN` to make it easy to spot "this is a function type" from its name, but I think it's just a matter of personal taste (i.e. does not help with avoiding name clash etc)

docs/developers/checklists.md
```diff
@@ -14,6 +14,7 @@ In preparation for a release, use the following checklist. These steps can be do
  - [ ] Change the version number in `Uno::current_version()` in `uno/Uno.cpp`
  - [ ] Change the version number in `CITATION.cff`
  - [ ] Change the version number in `CMakeLists.txt`
+ - [ ] Change the version number in `bindings/Uno_C_API.cpp`
```
amontoison: @cvanaret The version should be in `Uno_C_API.h` and `Uno_C_API.cpp` directly uses the version at compilation time.

People can check the version at compilation time in other projects (like C or Fortran) with the header or with the `uno_get_version` in high-level interfaces (like Julia or Python).

cvanaret: Fixed in 856580aabb5b5e3c0c5c65017f51b6c0a5fdc530.

amontoison:
@cvanaret I updated a little bit the C and Julia interfaces.
You can check the recent modifications with the latest commits.

Note that you forgot `objective_multiplier` in `uno_lagrangian_hessian_operator`.

cvanaret:
Thanks a lot @amontoison! I'll have a look.

The objective multiplier is already [baked in](https://github.com/cvanaret/Uno/pull/269/files#diff-7c251f6820e4f1e3d36ccd4cb4eb452aa2c0911f496daa1873374cf7e646115dR107) the `HessianOperator` type :)

cvanaret:
Neat! I haven't set a default preset in the C API. I should probably, and it gets overwritten if the user sets one.

cvanaret:
Default preset is set internally.
I'm not sure where `ERROR: LoadError: UndefVarError: uno_set_solver_preset not defined` comes from.

amontoison:
The function is not exported in the Julia interface `Uno.jl`.
I just exported the main one (`uno`).
We just need to prefix the other ones with `Uno.*`.

cvanaret:
Oops, I `push --force`d without rebasing, my bad!

bindings/C/Uno_C_API.cpp
```diff
@@ -351,8 +351,9 @@ void uno_set_lagrangian_hessian(void* model, int32_t number_hessian_nonzeros, ch
 
 void uno_set_lagrangian_hessian_operator(void* model, HessianOperator lagrangian_hessian_operator,
       double lagrangian_sign_convention) {
-   if (lagrangian_sign_convention != -UNO_MULTIPLIER_NEGATIVE && lagrangian_sign_convention != UNO_MULTIPLIER_POSITIVE) {
```
amontoison: @cvanaret Hum Hum `-UNO_MULTIPLIER_NEGATIVE`

cvanaret: Ahahah unchecked copy-paste is the devil.

cvanaret:
Well done @amontoison :beers:

cvanaret:
Weird, something broke and I'm not sure what. The C example still runs locally.
The only difference is the definition of `struct Solver` [here](https://github.com/cvanaret/Uno/blob/5301dc73dadd488a149aeea50422b9bdde09979d/bindings/C/Uno_C_API.cpp#L250). Otherwise, we've just added stuff to the API.

cvanaret:
There seems to be some kind of undefined behavior with the Jacobian sparsity when Uno switches to the restoration phase:
```
optimality phase:
C API Jacobian row indices: 0 0 1 1
Uno row indices:            0 0 1 1

restoration phase:
C API Jacobian row indices: 835152304 32756 835152352 32756
Uno row indices:            835152304 32756 835152352 32756
```

Also: more options should be printed before the optimization process, not just 3.
Not sure what's going on...

cvanaret:
OK, the hs015 example (`example_hs015.c`) runs fine (4 successive times) through the C API in CI: https://github.com/cvanaret/Uno/actions/runs/17101385493/job/48498630756?pr=269

This hints at a bug in the Julia interface. Not sure how Julia works, could it be an issue with temporaries? Perhaps the Uno model stores pointers to temporary vectors (the 0-based indexed Jacobian sparsity pattern, or `-multipliers` in `UnoNLPModelsExt.jl`), that exit the scope and we end up with dangling pointers and undefined behavior?

cvanaret:
Hi @amontoison,
I modified the getters to get the primal-dual solutions, [they now take a vector as second parameter](https://github.com/cvanaret/Uno/pull/269/commits/5e887b60ea1f1b9623de9442ab6e7c2a39643a98). I updated `C_wrapper.jl` and `libuno.jl`. Could you check what I did, and complete `runtests.jl`?

amontoison:
Thanks @cvanaret !! I wanted to ask you that yesterday and I forgot.
I will update `runtests.jl` a little bit later today.

amontoison:
> OK, the hs015 example (`example_hs015.c`) runs fine (4 successive times) through the C API in CI: https://github.com/cvanaret/Uno/actions/runs/17101385493/job/48498630756?pr=269
> 
> This hints at a bug in the Julia interface. Not sure how Julia works, could it be an issue with temporaries? Perhaps the Uno model stores pointers to temporary vectors (the 0-based indexed Jacobian sparsity pattern, or `-multipliers` in `UnoNLPModelsExt.jl`), that exit the scope and we end up with dangling pointers and undefined behavior?

`-multipliers` is only used in the Julia callbacks so it is not an issue. The value of `-multipliers` is not needed later because we do a new copy every-time until you support the sign `+1` of the Lagrange multipliers.

However if you don't do a copy of `jrows`, `jcols`, `hrows`, `hcols` in your C model after the `uno_create_model` then the Julia vectors can be garbage collected at any moment.
We can't access these vectors after the call to `uno` so Julia marks them as ready for GC.
If it is the case we need to protect then from GC and store them in the Julia structure `UnoModel`.

cvanaret:
Ah! That explains a lot.
Yes, let's copy them into the `UnoModel` 👍

amontoison:
Don't you want to have a copy of them in C/C++ such that it is always 0-based indexed?

cvanaret:
Yeah I could. The sparsity patterns are actually directly sent to the subproblem solver (BQPD, HSL, HiGHS, MUMPS, ...) with the right indexing, so in principle, having them with 0-based indexing is not crucial.
But that would probably simplify the development of a bunch of other interfaces as well (Matlab, C#, ...), so it's probably worth it :)

amontoison:
I am also scared that this issue with garbage collector arrises in other languages.

cvanaret:
Ok, looking good!

~I cannot explain why only 3 options are printed in the Julia test, though. In C/C++, 18 options are printed.~
Edit: fixed in https://github.com/cvanaret/Uno/commit/055b903e2b25d2f21c2907bbdb543d73c4499c8a.

What's left for me to do on the Uno side:
- [x] connect the Jacobian/Hessian operators from the C model to the Uno code.
- [x] handle the Lagrangian sign convention.
- [x] handle the triangle part of the Hessian.

amontoison:
For the triangle, I suggest to only work with the lower one internally and do that in the C interface:
```

**Comment by amontoison** [comment]
For the triangle of the Hessian of the Lagrangian, I suggest to only work with the lower triangle in Uno and do this "trick" in the C interface:
```c
   if (hessian_triangular_part == UNO_LOWER_TRIANGLE) {
      user_model->hessian_row_indices = hessian_row_indices;
      user_model->hessian_column_indices = hessian_column_indices;
   } else {
      user_model->hessian_row_indices = hessian_column_indices;
      user_model->hessian_column_indices = hessian_row_indices;
   }
```

amontoison:
On my side for the Julia interface `Uno.jl`, I need to add:
- A `README.md` that explains how to install the package, how to use `LBT`, `HSL_jll.jl`, etc...
- Write the JuMP interface (with a package extension for MOI)
- Write tutorials as documentation

amontoison:
@cvanaret https://github.com/cvanaret/Uno/tree/c_interface/bindings/Julia/Uno.jl

cvanaret:
> @cvanaret https://github.com/cvanaret/Uno/tree/c_interface/bindings/Julia/Uno.jl

Looks great, thanks :heart_eyes:

amontoison:
@cvanaret 
Can you have a look of the following output?
I don't understand why Uno complains for the objective sense.
Also Uno doesn't to solve the optimization problem if I provide the Jacobian / Hessian operators, even if they are not used.
```shell

Please specify a valid objective sense.

Used overwritten options:
- QP_solver = BQPD
- TR_min_radius = 1e-8
- TR_radius = 10
- dual_tolerance = 1e-6
- enforce_linear_constraints = yes
- funnel_update_strategy = 2
- l1_constraint_violation_coefficient = 1.
- primal_tolerance = 1e-6
- regularization_strategy = none
- residual_norm = L2

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 iter   TR iter  TR radius    phase  funnel width  step norm   objective   primal feas  stationarity  complementarity  status          
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 0      -        1.0000e+01   OPT    1.8000e+01    -           0.0000e+00  1.2000e+01   0.0000e+00    0.0000e+00       initial point   
 -      -        -            -      -             -           -           -            -             -                compute_hessian_vector_product not implemented yet
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 iter   TR iter  TR radius    phase  funnel width  step norm   objective   primal feas  stationarity  complementarity  status          
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Uno 2.0.3 (TR funnel restoration inequality-constrained method with exact Hessian and no regularization)
Tue Sep  9 06:01:21 2025
────────────────────────────────────────
Optimization status:			Algorithmic error
Iterate status:				Suboptimal point
Objective value:			0
Primal feasibility:			12
┌ Stationarity residual:		0
└ Complementarity residual:		0
┌ Infeasibility measure:		12
│ Objective measure:			0
└ Auxiliary measure:			0
Primal solution:			1 5 5 1 
┌ Constraint multipliers:		-0 -0 
│ Lower bound multipliers:		-0 -0 -0 -0 
└ Upper bound multipliers:		-0 -0 -0 -0 
Objective multiplier:			1
CPU time:				0.023346s
Iterations:				1
Objective evaluations:			1
Constraints evaluations:		1
Objective gradient evaluations:		1
Jacobian evaluations:			0
Hessian evaluations:			0
Number of subproblems solved:		0
```

cvanaret:
> @cvanaret Can you have a look of the following output? I don't understand why Uno complains for the objective sense. Also Uno doesn't to solve the optimization problem if I provide the Jacobian / Hessian operators, even if they are not used.

`uno_set_objective` (and similar functions) now return a boolean that can be tested (false when failed). I also fixed the hard-coded constants in the Julia interface (TO-DO: load the C constants in Julia @amontoison).

By default, BQPD tries to use Hessian-vector products when possible. The Julia interface passes one, so Uno tried to use it but the function in the C interface was not implemented :) That's fixed too.

There is still a bug upon evaluation of the Hessian operator during CI, I need to investigate that. I disabled the Hessian operators, and the CI test runs with explicit Hessian matrices :+1:

cvanaret:
@amontoison we're good to go :sunglasses:

cvanaret:
Great work @amontoison! :star_struck:
The next step (finalizing and testing the MOI/JuMP interface) will be performed in a separate PR.

PR #278, opened 2025-08-15
Compile-time switch between int32_t and int64_t

cvanaret:
- [x] add a compile-time switch between `int32_t` and `int64_t`
- [ ] use the resulting `UnoInt` in the code (indices for sparsity patterns)

cc @amontoison

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,11 @@ option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
 if(NOT BUILD_STATIC_LIBS AND NOT BUILD_SHARED_LIBS)
    message(FATAL_ERROR "At least one of BUILD_SHARED_LIBS or BUILD_STATIC_LIBS must be ON.")
 endif()
+add_compile_definitions()
+option(USE_64BIT_INTEGERS "Use 64-bit integers" OFF)
+if(USE_64BIT_INTEGERS)
+   add_compile_definitions(USE_64BIT_INTEGERS)
+endif()
 
 # determine whether a Fortran compiler is required, based on the available optional dependencies
 find_library(HSL hsl)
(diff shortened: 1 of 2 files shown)


amontoison:
@cvanaret I propose to not support yet the `int64_t` but use the `UnoInt` everywhere such that you don't add more code where your hardcode some `int32_t`.
If one day someone request this feature (support of `int64_t`), it will be easier to add it.
It was an advice for making the code more modular ;)

cvanaret:
> @cvanaret I propose to not support yet the `int64_t` but use the `UnoInt` everywhere such that you don't add more code where your hardcode some `int32_t`. If one day someone request this feature (support of `int64_t`), it will be easier to add it. It was an advice for making the code more modular ;)

Sounds good!

cvanaret:
@amontoison done!
I duplicated the definition `uno_int = int32_t` in `tools/uno_int.hpp` and `interfaces/C/Uno_C_API.h`, don't think it's too dramatic?
Also, I'm not sure how to define type aliases in Fortran so I hardcoded `c_int32_t` in `interfaces/Fortran/[test_]uno.f90`. I'm sure we can do something better!

amontoison:
> @amontoison done! I duplicated the definition `uno_int = int32_t` in `tools/uno_int.hpp` and `interfaces/C/Uno_C_API.h`, don't think it's too dramatic?

It is not dramatic but is it not possible to do a `uno_int.h` instead of a `uno_int.hpp` that you can include in both C and C++ ?

> Also, I'm not sure how to define type aliases in Fortran so I hardcoded `c_int32_t` in `interfaces/Fortran/[test_]uno.f90`. I'm sure we can do something better!

I will have a look 🧐

imciner2:
> > Also, I'm not sure how to define type aliases in Fortran so I hardcoded c_int32_t in interfaces/Fortran/[test_]uno.f90. I'm sure we can do something better!
>
> I will have a look 🧐

You can define a new type kind to use in the various integer definitions just by doing this:

```fortran
integer, parameter :: uno_int = c_int32_t
```

cvanaret:
Thanks @imciner2 and @amontoison!

PR #331, opened 2025-10-06
[C interface] Copy internally the vectors passed by the user

cvanaret:
Do not store user pointers! Always make copies.
- C interface:
  - loops instead of calling `std::copy_n` (which moves the pointer)
  - makes copies of the vectors
- unopy: fixed the implementation of `Model.cpp`

cc @amontoison

### DIFF
--- a/interfaces/C/Uno_C_API.cpp
+++ b/interfaces/C/Uno_C_API.cpp
@@ -480,24 +480,30 @@ bool uno_set_user_data(void* model, void* user_data) {
    return true;
 }
 
-bool uno_set_initial_primal_iterate(void* model, double* initial_primal_iterate) {
+bool uno_set_initial_primal_iterate(void* model, const double* initial_primal_iterate) {
    assert(model != nullptr);
-   CUserModel* user_model = static_cast<CUserModel*>(model);
-   // copy the initial primal point
-   for (size_t variable_index: Range(static_cast<size_t>(user_model->number_variables))) {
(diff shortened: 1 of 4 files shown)


amontoison (approved):

interfaces/C/Uno_C_API.cpp
```diff
@@ -96,31 +96,35 @@ class UnoModel: public Model {
...
          }
       }
    }
 
    void compute_hessian_sparsity(int* row_indices, int* column_indices, int solver_indexing) const override {
       // copy the indices of the user sparsity patterns to the Uno vectors
-      std::copy_n(this->user_model.hessian_row_indices.data(), static_cast<size_t>(this->user_model.number_hessian_nonzeros), row_indices);
-      std::copy_n(this->user_model.hessian_column_indices.data(), static_cast<size_t>(this->user_model.number_hessian_nonzeros), column_indices);
+      for (size_t nonzero_index: Range(static_cast<size_t>(this->user_model.number_hessian_nonzeros))) {
+         row_indices[nonzero_index] = this->user_model.hessian_row_indices[nonzero_index];
```
amontoison, line 118: @cvanaret The copy should be in the `this->user_model.hessian_row_indices` and not the contrary Charlie?!

cvanaret: @amontoison I thought so for a while, but this is just Uno asking the model for the sparsity patterns and copying them into the vectors in parameters. So all good :)

Issue #332, opened 2025-10-06
The C model is shared between different Julia models ☢️


amontoison:
```julia
alexis@HP-Spectre:~/Téléchargements$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.11.7 (2025-09-08)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> include("uno.jl")
Original model C model
2 variables, 1 constraints (0 equality, 1 inequality)

Used overwritten options:
- QP_solver = BQPD
- TR_min_radius = 1e-8
- TR_radius = 10
- dual_tolerance = 1e-6
- enforce_linear_constraints = yes
- funnel_update_strategy = 2
- l1_constraint_violation_coefficient = 1.
- primal_tolerance = 1e-6
- regularization_strategy = none
- residual_norm = L2

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 iter   TR iter  TR radius    phase  funnel width  step norm   objective   primal feas  stationarity  complementarity  status          
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 0      -        1.0000e+01   OPT    1.0000e+00    -           -0.0000e+00 0.0000e+00   1.0000e+00    0.0000e+00       initial point   
 1      1        1.0000e+01   OPT    1.0000e+00    1.0000e+00  -1.0000e+00 3.0685e-01   0.0000e+00    3.0685e-01       ✔ (f-type)      
 2      1        1.0000e+01   OPT    5.0000e-01    3.0685e-01  -6.9315e-01 6.5982e-310  0.0000e+00    0.0000e+00       ✔ (h-type)      
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 iter   TR iter  TR radius    phase  funnel width  step norm   objective   primal feas  stationarity  complementarity  status          
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Uno 2.2.0 (TR funnel restoration inequality-constrained method with exact Hessian and no regularization)
Mon Oct  6 12:53:13 2025
────────────────────────────────────────
Optimization status:			Success
[...]

PR #335, opened 2025-10-07
Update the C interface to set and get only components

amontoison:
- We don't need to allocate additional vectors with these new C routines.


cvanaret:
@amontoison :v: 
But the function evaluation count accumulates as you reported in https://github.com/cvanaret/Uno/issues/332.

PR #336, opened 2025-10-07
Added missing result getters in C and Julia interfaces

cvanaret:


amontoison (approved):

amontoison:
@cvanaret Do you want to rename the `setter`  to be consistent?
Currently we have:
- `uno_set_initial_primal_component`
- `uno_set_initial_dual_component`
- `uno_set_initial_primal_iterate`
- `uno_set_initial_dual_iterate`

cvanaret:
@amontoison good point. I'm happy with how it is at the moment (`initial` is an adjective so is in front, `solution` is a noun so in the end).

amontoison:
> @amontoison good point. I'm happy with how it is at the moment (`initial` is an adjective so is in front, `solution` is a noun so in the end).

So you don't want `..._iterate_component` like what you did for the `getters`?

cvanaret:
> > @amontoison good point. I'm happy with how it is at the moment (`initial` is an adjective so is in front, `solution` is a noun so in the end).
> 
> So you don't want `..._iterate_component` like what you did for the `getters`?

Ah sorry, I misread, I thought you suggested moving the `initial` to the end. Yes for the `_component` functions we can add `_iterate` before `_component` :+1: 
Addressed in https://github.com/cvanaret/Uno/pull/339.

PR #341, opened 2025-10-08
Typed functions to set solver options

cvanaret:
- [x] added typed functions
  - `Options::set_integer`
  - `Options::set_double`
  - `Options::set_bool`
  - `Options::set_string`
- [x] added the functions to the C and Julia interfaces
- [x] fix the tests of `Uno.jl`
- [x] handled the command line arguments for `uno_ampl`
- [x] fix the tests of `Uno_jll.jl`
- [x] handle the preset in `uno_set_solver_string_option()` as well

### DIFF
--- a/interfaces/Julia/ext/UnoMathOptInterfaceExt/MOI_wrapper.jl
+++ b/interfaces/Julia/ext/UnoMathOptInterfaceExt/MOI_wrapper.jl
@@ -1432,11 +1432,7 @@ function MOI.optimize!(model::Optimizer)
     # Other misc options that over-ride the ones set above.
     for (name, value) in model.options
         if value isa String
-            if name == "preset"
-                Uno.uno_set_solver_preset(solver, value)
-            else
-                Uno.uno_set_solver_string_option(solver, name, value)
-            end
+            Uno.uno_set_solver_string_option(solver, name, value)
         elseif value isa Float64
             Uno.uno_set_solver_double_option(solver, name, value)
         elseif value isa Integer


cvanaret:
@amontoison I'm not completely done yet (I need to handle command line arguments for `uno_ampl`), but could you have a look at the changes to `Uno.jl`?

interfaces/C/Uno_C_API.cpp
```diff
@@ -676,9 +670,24 @@ void* uno_create_solver() {
...
+   Solver* uno_solver = static_cast<Solver*>(solver);
+   uno_solver->options->set_integer(option_name, option_value);
+}
+
+void uno_set_solver_double_option(void* solver, const char* option_name, double option_value) {
+   Solver* uno_solver = static_cast<Solver*>(solver);
+   uno_solver->options->set_double(option_name, option_value);
+}
+
+void uno_set_solver_bool_option(void* solver, const char* option_name, bool option_value) {
```
amontoison, line 690: @cvanaret I think we should merge it with `uno_set_solver_integer_option` and just check that for these boolean arguments, the user passed only `0` or `1`.

amontoison: In the time, it is more clean with a `bool`...
Let's keep what your propose and see if Fortran users complain in the future :)

amontoison:
@cvanaret 
I did a pass on your modifications: https://github.com/cvanaret/Uno/pull/341/commits/cef6570d1ccae85571160e9f28b59156baeb9145

It looks good :+1: 
It will be great to remove the routine `uno_set_solver_preset` and only relies on `uno_set_solver_string_option`.

cvanaret:
@amontoison great, thank you!
I'm a little bit hesitant to handle presets as regular options, since a preset is more on a "meta" level (as a collection of options). But in the end, it probably doesn't matter too much I guess. After all, the command line syntax for presets and options is the same for `uno_ampl`.

amontoison:
If you want to keep it, can you also support to set the preset with` set_string_option`?
This way, the user can choose what he wants.
For the Julia interface, we can avoid to do special cases for the option name (`"preset"`) this way.

cvanaret:
@amontoison we're good to go!
The preset is now handled in the C interface (and not in `MOI_wrapper.jl` anymore) by:
- `uno_set_solver_preset(void* solver, const char* preset_name)` or
- `uno_set_solver_string_option(void* solver, const char* option_name, const char* option_value)`

PR #350, opened 2025-10-11
Allow stopping UNO from user callbacks

stefphd:
Allow users to stop UNO during optimization through user callbacks.

- New `SolutionStatus ::USER_REQUESTED_STOP` and corresponding status message
- New method `void UserCallbacks::stop_request()` to request stopping UNO. This sets the internal flag `UserCallbacks::stop_request_flag`. Marked as `virtual final` to prevent overriding
- Added supporting methods: `bool UserCallbacks::is_stop_requested()` and `void UserCallbacks::reset_stop_request()`
- The stop requested flag is checked in `Uno::uno_solve` before evaluating the termination criteria
- Updated the C interface: user callbacks now return a `bool` (`false` indicates a stop request)

In C++ I preferred this implementation (explicit `stop_request()` and flag) over using a callback that returns a `bool`. Propagating the iteration status through return values was tricky and error-prone, especially for `notify_acceptable_iterate`.

Personally I sometime use this feature. Let me know if you have any suggestions or improvements. Feel free to accept or reject if you think it’s not worth including.

### DIFF
--- a/interfaces/Julia/gen/prologue.jl
+++ b/interfaces/Julia/gen/prologue.jl
@@ -32,7 +32,7 @@ const UNO_ITERATION_LIMIT = Cint(1)
 const UNO_TIME_LIMIT = Cint(2)
 const UNO_EVALUATION_ERROR = Cint(3)
 const UNO_ALGORITHMIC_ERROR = Cint(4)
-const UNO_USER_REQUESTED_STOP = Cint(5)
+const UNO_USER_TERMINATION = Cint(5)
 
 # Iterate status
 const UNO_NOT_OPTIMAL = Cint(0)
--- a/interfaces/Julia/gen/wrapper.jl
+++ b/interfaces/Julia/gen/wrapper.jl
@@ -17,7 +17,8 @@ function main()
                "JacobianOperator", "JacobianTransposedOperator", "HessianOperator",
                "JacobianSparsity", "Jacobian", "HessianSparsity", "Hessian",
                "NotifyAcceptableIterateUserCallback", "NotifyNewPrimalsUserCallback",
-               "NotifyNewMultipliersUserCallback", "LoggerStreamUserCallback"]
+               "NotifyNewMultipliersUserCallback", "TerminationUserCallback",
+               "LoggerStreamUserCallback"]
   options["general"]["output_ignorelist"] = callbacks
 
   args = get_default_args()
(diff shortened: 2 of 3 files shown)


cvanaret:
Thanks @stefphd, it's definitely worth adding this feature.
I'll have a closer look tomorrow/on Monday.

My first comment would be that `USER_REQUESTED_STOP` should be a value of the `OptimizationStatus` enum (like `TIME_LIMIT` for example), not `SolutionStatus`. Upon user-requested stop, the `solution_status` then tells you what the final iterate looks like (feasible? stationary?). Do you agree?

stefphd:
Hi Charlie, 

It sounds better to me: changes committed.

I have also changed `Uno::termination_criteria` to use the same workflow (basically, set `optimization_status` in `Uno::termination_criteria`)

stefphd:
- added `OptionType Options::get_option_type(const std::string& option_name)`
- added  `int32_t uno_get_solver_option_type(void* solver, const char* option_name)` and corrisponding C option types `UNO_OPTION_TYPE_INT, UNO_OPTION_TYPE_DOUBLE` etc.
- made option setter and getter names consistent in C interface
- re-generated julia wrapper
- updated C interface readme

Cf. #341 

Mess in commit [439c15f](https://github.com/stefphd/Uno/commit/439c15fb779395e46c5ef5027cf819e9e11da876) (reverted and fixed with [f864307](https://github.com/cvanaret/Uno/pull/350/commits/f8643078dc79ba349482e9b047aa1ced285ca4a0) and [fa7a0bb](https://github.com/cvanaret/Uno/pull/350/commits/fa7a0bbd611ca99831b3edef12e8dfbeec9af5d3))

cvanaret:
Hi @stefphd,
Sorry, I was busy with all sorts of things and I'm starting only now to review your PR.

We definitely need this feature in Uno. I would suggest either:
1. having the callbacks return an enum value (say, `CONTINUE` or `STOP`) instead of a boolean, or
2. (my preferred option) introducing a callback specifically for termination:
    `bool user_termination_callback(primal_iterate, dual_iterate, residuals, ...)` 
that would be called for example [here](https://github.com/cvanaret/Uno/blob/main/uno/Uno.cpp#L188).

As is, my concern with the suggested design is that the callbacks would now perform two distinct tasks (notify the user with the new iterate + test termination) in one single function, even though the names of the callbacks do not indicate so. I'd be happier with a better “separation of concerns” (solution 2 above).
Of course open for discussion :slightly_smiling_face:

stefphd:
Hi Charlie,

For solution 1, looking around the code it seems a bit tricky and error-prone to propagate the stop request. On the other side, this is similar to what's done in other solvers (eg ipopt, although only a single callback is used there). 

I do think that option 2 is a very good and clean idea. I will definitely try to implement this. I would expect that this callback takes all input arguments of the notify accepted iterate callback.

cvanaret:
This looks great! I'm merging into main. Thanks Stefano!
I'll make a few tiny changes in a separate PR.

PR #353, opened 2025-10-14
[Uno.jl] Support the silent mode in the MOI extension

amontoison:


amontoison:
@cvanaret I think I found a bug in `Uno` by not imposing any `preset` by default.

cvanaret:
@amontoison oh looks like I forgot to implement the same behavior as in `uno_ampl`! I'll adress that this evening 👍

amontoison:
@cvanaret I added a comment for you in the code about an "acceptable" primal tolerance.
All tests passed a rebase on top of #354 :+1:

cvanaret:
> @cvanaret I added a comment for you in the code about an "acceptable" primal tolerance

That's a good idea! I'll open an issue.

PR #360, opened 2025-10-15
[WIP] MATLAB interface

stefphd:
Hi Charlie,

I've just written a first draft of the MATLAB interface. It has also some minor changes in the C/C++ interface. For now it's only a draft, may be unstable at the moment.

Track list:
- [x] Create low-level matlab mex functions `result = uno_optimize(model[, options, callbacks])` and `options = uno_options([preset])`
- [x] Create readme for matlab interface
- [x] Create HS015 problem example
- [x] Use explicit sparsity pattern in matlab low-level interface
- [x] Add argument validation functions in `interfaces\Matlab\Uno_mex_utilities.hpp`
- [x] Add argument validation in`uno_optimize` mex function
- [x] Add high-level (fmicon-like) interface
- [x] Add github action (Windows)
- [x] Add github action (MacOS)


cvanaret:
Really cool Stefano!
Do we need a self-hosted runner with a Matlab license to test the interface in a Github Action or can we do that on the Github runners?

stefphd:
Hi @cvanaret 

As far I know, it's possible to use Matlab on the Github runners for public repos, so no need for self-hosted runners (see [here](https://github.com/matlab-actions/.github?tab=readme-ov-file#use-matlab-batch-licensing-token)). So far, I was able to build a Matlab mex (via cmake) even for a private repo without a Matlab license, as this step only requires the Matlab installation and doesn’t actually call MATLAB itself. I’ll try running a test on GitHub runners to confirm that this works as well.

stefphd:
Tested Matlab mex in Github runner: [matlab-action-test](https://github.com/stefphd/matlab-action-test)
No Matlab license needed

For Mac-os a workaround seems to be needed when building the mex using CMake: add the (typical) location(s) of `libMatlabEngine.dylib` to the mex rpath: 

```cmake
if(APPLE)
    # This adds $matlabdir/extern/bin/maci64 (intel arch) and $matlabdir/extern/bin/maca64 (arm arch) to @rpath
    # @executable_path is the location of the executable, i.e. matlab, typically $matlabdir/bin or $matlabdir/bin/<arch>
    # This is also to solve bug in MATLAB R2023b, see 
    # https://it.mathworks.com/matlabcentral/answers/2044762-why-do-i-receive-a-library-not-loaded-rpath-libmatlabengine-dylib-error-when-running-a-mex-funct  
	set(MEX_RPATH "@executable_path/../extern/bin/maci64;@executable_path/../extern/bin/maca64;@executable_path/../../extern/bin/maci64;@executable_path/../../extern/bin/maca64")
	set_target_properties(mex-target PROPERTIES 
                BUILD_RPATH "${CMAKE_BUILD_RPATH};${MEX_RPATH}"
                INSTALL_RPATH "${CMAKE_INSTALL_RPATH};${MEX_RPATH}"
        )
endif()
```

interfaces/Matlab/README.md
```diff
@@ -1,12 +1,114 @@
...
+- high-level fmincon-style interface through the Matlab function `uno`
+- low-level interface through the Matlab MEX functions `uno_optimize` and `uno_options`; the two functions are compiled via the command `make unomex`.
 
-Uno's MATLAB interface allows you to solve an optimization model described in MATLAB.
-The MATLAB MEX functions `uno_optimize` and `uno_options` are compiled via the command `make unomex`.
 An example is available in the file [example_hs015.m](example/example_hs015.m).
 
+## Affiliation
+
+This Julia interface is developed and maintained by [Stefano Lovato](https://github.com/stefphd) and [Charlie Vanaret](https://github.com/cvanaret).
```
cvanaret: The **Matlab** interface :)

interfaces/Matlab/README.md
```diff
@@ -1,12 +1,114 @@
...
+
+Define the objective function:
+
+```matlab
+function [fval,grad,hess] = fun(x)
+    % ...
+end
+```
+
+- the objetive value `fval` must be a scalar numeric value;
```
cvanaret: obje**c**tive

interfaces/Matlab/README.md
```diff
@@ -11,7 +11,7 @@ An example is available in the file [example_hs015.m](example/example_hs015.m).
 
 ## Affiliation
 
-This Julia interface is developed and maintained by [Stefano Lovato](https://github.com/stefphd) and [Charlie Vanaret](https://github.com/cvanaret).
+This Objective interface is developed and maintained by [Stefano Lovato](https://github.com/stefphd) and [Charlie Vanaret](https://github.com/cvanaret).
```
cvanaret: Still not :smile:

stefphd: Oh god, my brain is fried

cvanaret:
The Github action runs, well done :partying_face:

cvanaret:
@stefphd sorry for the many changes we're merging into main :sweat_smile: Let me know if you need anything.

CMakeLists.txt
```diff
@@ -297,6 +297,58 @@ if(pybind11_DIR)
    target_include_directories(unopy PUBLIC ${DIRECTORIES})
 endif()
 
+############################
+# optional Matlab bindings #
+############################
+
+find_package(Matlab COMPONENTS MAIN_PROGRAM MEX_COMPILER)
```
worc4021: If you've verified this works with a particular matlab version I'd recommend pinning it on that version. There have been several minor changes in the mex behaviour over the past couple of versions that could cause unnecessary heartache.

cvanaret:
Hi @stefphd, can I help with something that is blocking the further development of the Matlab interface? Or is it simply lack of time?

stefphd:
Hi @cvanaret

I was waiting for the L-BFGS feature in IPM, which is very useful. Actually, I see you've just released UNO including L-BFGS for IPM, that's very nice!
I plan to finalize the Matlab interface in the next weeks.

cvanaret:
> Hi @cvanaret
> 
> I was waiting for the L-BFGS feature in IPM, which is very useful. Actually, I see you've just released UNO including L-BFGS for IPM, that's very nice! I plan to finalize the Matlab interface in the next weeks.

Great timing then :)

CMakeLists.txt
stefphd: Verified up to release R2023b (corresponding to version 23.2). 
Now minimum required version is 23.2, see commit [6313475](https://github.com/cvanaret/Uno/pull/360/commits/631347569b8345c1c9fca9116b926bf08c8fa506).

stefphd:
Matlab interface is almost completed; a few tests may still be required.

Currently:
- The high-level interface should be quite robust and stable (it mimics the Matlab fmincon command, with a few differences explained in the README)
- The low-level interface is more risky (an incorrect problem setup may cause Matlab to crash with a segmentation fault, etc.), although it provides complete control over Uno. Under the hood, the high-level interface employs the low-level interface.

I was able to successfully run the MATLAB test examples on both Windows and macOS GitHub runners; however, some Python tests are still failing (I don’t know the reason yet).

Note that macOS was quite challenging: a segfault during cleanup on exit causes MATLAB to crash. The current workaround is to lock the compiled MEX file to prevent it from being unloaded, so that TLS cleanup handlers remain valid (cf. [matlab_openmp_tls_bug](https://github.com/wmutschl/matlab_openmp_tls_bug)).

cvanaret:
Great job! I'll have a look why the Python workflows fails.

cvanaret:
OK it seems to work. The bug had nothing to do with the Matlab interface but was apparently caused by the class `UserStreamCallback` being moved to the unopy files (for which LTO was enabled).
I'll do a separate fix which you can then merge.

stefphd:
Thanks Charlie!!!
I created a base class `UserStreamCallback` for a common interface to all stream callbacks (`CStreamCallback` and `MatlabStreamCallback`) to avoid duplicated code.

Edit: I see that the Python interface uses a slightly different approach. I'm still not sure which one is best.

cvanaret:
It's ok if the Python and Matlab interfaces don't use exactly the same approach. We should "ship fast" and improve later :)

PR #373, opened 2025-10-22
Rename Uno.jl into UnoSolver.jl

amontoison:


interfaces/Julia/README.md
```diff
@@ -77,7 +77,7 @@ objective_value(jump_model)     # objective value
 value.(x)                       # primal solution
 ```
 
-If you encounter any issues with the interface for JuMP problems, please [open an issue](https://github.com/JuliaSmoothOptimizers/Uno.jl/issues) so we can fix it.
+If you encounter any issues with the interface for JuMP problems, please [open an issue](https://github.com/JuliaSmoothOptimizers/UnoSolver.jl/issues) so we can fix it.
```
amontoison:
```suggestion
If you encounter any issues with the interface for JuMP problems, please [open an issue](https://github.com/cvanaret/Uno/issues) so we can fix it.
```

cvanaret:
Thanks @amontoison! Is `using UnoSolver` missing [here](https://github.com/cvanaret/Uno/blob/14ce7fa0164935decfe39a2b382c7d63223689e1/interfaces/Julia/README.md?plain=1#L84)?

amontoison:
> Thanks @amontoison! Is `using UnoSolver` missing [here](https://github.com/cvanaret/Uno/blob/14ce7fa0164935decfe39a2b382c7d63223689e1/interfaces/Julia/README.md?plain=1#L84)?

I was expecting the user to call this block after the previous one where we load `UnoSolver` and `JuMP`.
But let's add it, it is free.

PR #376, opened 2025-10-23
[UnoSolver.jl] Don't use operators if we have a VectorNonlinearOracle in the MOI model

amontoison:
@odow Can you confirm it?

### DIFF
--- a/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
+++ b/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
@@ -1599,7 +1599,7 @@ end
 function MOI.get(
     model::Optimizer,
     attr::MOI.ConstraintPrimal,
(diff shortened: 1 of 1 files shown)


odow:
I don't understand the question. What do operators have to do with the oracle?

amontoison:
If the user uses an oracle you can only compute the sparse Jacobian / sparse Hessian of the Lagrangian, but some nonlinear solvers like `Uno` or `KNITRO` has algorithms that only require operators `H * v` where `H` is the Hessian of the Lagrangian.

In this PR, I force Uno (and internally BQPD) to not use Hess-vec products because we can't easily recover it from an oracle.

It is related to this comment: https://github.com/jump-dev/Ipopt.jl/issues/468#issuecomment-3350864442

odow:
Oh, you mean hess-vec operator, not `JuMP.@operator`. 

I missed the request for hess-vec, but I think it's out of scope. The set is already complicated; let's just stick with hessian.

amontoison:
Ok, I just wanted a confirmation.
Thanks Oscar :+1:

PR #379, opened 2025-10-23
Make problem_type a string/const char* instead of a char

cvanaret:


interfaces/Julia/gen/prologue.jl
```diff
@@ -6,10 +6,10 @@ const UNO_MAXIMIZE = Cint(-1)
 const UNO_MULTIPLIER_POSITIVE = Cdouble(1.0)
 const UNO_MULTIPLIER_NEGATIVE = Cdouble(-1.0)
 
-# Problem type: 'L' = Linear, 'Q' = Quadratic, 'N' = Nonlinear
-const UNO_PROBLEM_LINEAR    = Cchar('L')
-const UNO_PROBLEM_QUADRATIC = Cchar('Q')
-const UNO_PROBLEM_NONLINEAR = Cchar('N')
+# Problem type: "LP" = linear problem, "QP" = quadratic problem, "NLP" = nonlinear problem
+const UNO_PROBLEM_LINEAR    = Cstring("LP")
```
cvanaret: @amontoison I need your expertise here. I get:
```julia
ERROR: LoadError: MethodError: no method matching Cstring(::String)
Closest candidates are:
  Cstring(::Union{Ptr{Int8}, Ptr{Nothing}, Ptr{UInt8}})
```

amontoison: You can keep the `string` here, Julia will convert them into `Cstring` during the `@ccall` in `libuno.jl`.
Just for your knowledge, Julia will do that internally during the `@ccall`:
```shell
Base.unsafe_convert(Cstring, "LP")
```
We don't want to do directly in `prologue.jl` / `libuno.jl` because it returns an address of a vector of `UInt8` (vector of characters) and may be a pain to handle with the Julia garbage collector.
It is handled automatically when the conversion is done in `@ccall`.

cvanaret: Cool, I'll give that a try, thanks!

amontoison: From a low-level approach, the C interface wants the pointer of the following array `[Cchar('L'), Cchar('P'), Cchar('\0')]`. Julia will take care of doing it but it is what we need to do by hand in the Fortran interface.

cvanaret: Sounds fun :grinning:

amontoison (approved):

PR #380, opened 2025-10-23
Have setters (options, preset, callbacks, stream) return a boolean

cvanaret:


amontoison (changes requested):
We also need to update the Fortran interface (`uno.f90`).

PR #391, opened 2025-10-25
Postprocess the signs of the constraint multipliers consistently

cvanaret:
Postprocess the signs of the optimal constraint multipliers consistently in `Uno.cpp`, instead of in the individual `Model` subclasses:
- flip sign if (minimization + positive Lagrangian convention) or (maximization + negative Lagrangian convention)
- don't flip otherwise.

`AMPLModel` is a special case because we can set the same Lagrangian convention as Uno (`UNO_MULTIPLIER_NEGATIVE`), but it is ignored when we write the solution into a file (`write_solution`) :unamused:

### DIFF
--- a/uno/Uno.cpp
+++ b/uno/Uno.cpp
@@ -236,8 +236,10 @@ namespace uno {
          // do nothing
       }
       else {
-         // change the signs of the constraint multipliers
+         // change the signs of the multipliers
          result.constraint_dual_solution.scale(-1.);
+         result.lower_bound_dual_solution.scale(-1.);
+         result.upper_bound_dual_solution.scale(-1.);
       }
       result.solution_objective *= model.optimization_sense;
    }


stefphd:
Should [PythonModel::postprocess_solution](https://github.com/cvanaret/Uno/blob/6dd07f8d0a8d8ceb3cbb691b33d48ef69fecf40f/interfaces/Python/cpp_classes/PythonModel.cpp#L277) do nothing? See also [‎UnoModel::postprocess_solution](https://github.com/cvanaret/Uno/blob/6dd07f8d0a8d8ceb3cbb691b33d48ef69fecf40f/interfaces/C/Uno_C_API.cpp#L284)

cvanaret:
@stefphd absolutely, that one slipped through the net. Thanks!
Fixed in https://github.com/cvanaret/Uno/pull/396

PR #394, opened 2025-10-27
[UnoSolver.jl] fix dual convention and associated tests

odow:
@amontoison there are still some failing tests that I need to debug.

### DIFF
--- a/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
+++ b/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
@@ -747,30 +747,18 @@ function MOI.get(
 ) where {F<:MOI.VectorOfVariables,S<:MOI.VectorNonlinearOracle{Float64}}
     MOI.check_result_index_bounds(model, attr)
     MOI.throw_if_not_valid(model, ci)
-    sign = -_dual_multiplier(model)
     f, s = model.vector_nonlinear_oracle_constraints[ci.value]
-
-    # Alexis -- revisit this code when we know how to handle the workspace
-    # λ = model.inner.mult_g[row(model, ci)]
-    mult_g = Vector{Float64}(undef, model.inner.ncon)
-    UnoSolver.uno_get_constraint_dual_solution(model.solver, mult_g)
-
-    λ = mult_g[row(model, ci)]
-    J = Tuple{Int,Int}[]
-    _jacobian_structure(J, 0, f, s)
-    J_val = zeros(length(J))
-
-    # Alexis -- revisit this code when we know how to handle the workspace
-    # _eval_constraint_jacobian(J_val, 0, model.inner.x, f, s)
-    sol = Vector{Float64}(undef, model.inner.nvar)
-    UnoSolver.uno_get_primal_solution(model.solver, sol)
-    _eval_constraint_jacobian(J_val, 0, sol, f, s)
-
+    J = zeros(length(s.set.jacobian_structure))
+    x = [MOI.get(model, MOI.VariablePrimal(), xi) for xi in f.variables]
+    s.set.eval_jacobian(J, x)
+    λ = Float64[
+        UnoSolver.uno_get_constraint_dual_solution_component(model.solver, r - 1)
+        for r in row(model, ci)
+    ]
     dual = zeros(MOI.dimension(s.set))
-    # dual = λ' * J(x)
-    col_to_index = Dict(x.value => j for (j, x) in enumerate(f.variables))
-    for ((row, col), J_rc) in zip(J, J_val)
-        dual[col_to_index[col]] += sign * J_rc * λ[row]
+    sign = _dual_multiplier(model)
+    for ((row, col), J_rc) in zip(s.set.jacobian_structure, J)
+        dual[col] += sign * λ[row] * J_rc
     end
     return dual
 end
@@ -1649,8 +1637,7 @@ end
 ### MOI.ConstraintDual
 
 function _dual_multiplier(model::Optimizer)
-    sign = model.problem_type == "LP" ? 1.0 : -1.0
-    return model.sense == MOI.MIN_SENSE ? sign : -sign
+    return xor(model.problem_type == "LP", model.sense == MOI.MAX_SENSE) ? 1.0 : -1.0
 end
 
 function MOI.get(
--- a/interfaces/Julia/test/MOI_wrapper.jl
+++ b/interfaces/Julia/test/MOI_wrapper.jl
@@ -844,16 +844,16 @@ function test_vector_nonlinear_oracle_optimization()
     @test MOI.get(model, MOI.TerminationStatus()) == MOI.LOCALLY_SOLVED
     @test MOI.get(model, MOI.PrimalStatus()) == MOI.FEASIBLE_POINT
     @test MOI.get(model, MOI.DualStatus()) == MOI.FEASIBLE_POINT
-    atol = 1e-6
+    atol = 1e-3
     x_sol = MOI.get.(model, MOI.VariablePrimal(), x)
     @test ≈(x_sol, [1 / sqrt(2), 1 / sqrt(2), 1.0, 0.0]; atol)
     @test ≈(MOI.get(model, MOI.VariablePrimal(), t), -log(sqrt(2)); atol)
     c_sol = MOI.get(model, MOI.ConstraintPrimal(), c)
     @test ≈(c_sol, [1 / sqrt(2), 1 / sqrt(2), 1.0, 0.0]; atol)
     c_dual = MOI.get(model, MOI.ConstraintDual(), c)
-    @test ≈(c_dual, [-sqrt(2), 0.0, 0.5, -1 / sqrt(2)]; atol = 1000 * atol)  # <-- Relax the absolute tolerance
+    @test ≈(c_dual, [-sqrt(2), 0.0, 0.5, -1 / sqrt(2)]; atol)
     @test ≈(MOI.get(model, MOI.ConstraintDual(), c_x3), -0.5; atol)
-    @test ≈(MOI.get(model, MOI.ConstraintDual(), c_x4), 1 / sqrt(2); atol = 1000 * atol)  # <-- Relax the absolute tolerance
+    @test ≈(MOI.get(model, MOI.ConstraintDual(), c_x4), 1 / sqrt(2); atol)
     @test ≈(MOI.get(model, MOI.ConstraintPrimal(), c_snf), 0.0; atol)
     @test ≈(MOI.get(model, MOI.ConstraintDual(), c_snf), 1.0; atol)
     return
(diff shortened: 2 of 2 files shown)


interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1656,10 +1656,19 @@ function MOI.get(
...
+
+function MOI.get(
+    model::Optimizer,
+    attr::MOI.ConstraintDual,
+    ci::MOI.ConstraintIndex{MOI.ScalarQuadraticFunction{Float64},<:_SETS},
+)
+    MOI.check_result_index_bounds(model, attr)
+    MOI.throw_if_not_valid(model, ci)
+    λ = UnoSolver.uno_get_constraint_dual_solution_component(model.solver, row(model, ci)-1)
+    return -_dual_multiplier(model) * λ
```
amontoison: Is it a duplicate of the previous function?

odow: Nope. It has a `-_dual`.

odow:
I think I understand the crux of the matter. The dual convention seems to depend also on whether the objective is linear or quadratic

odow:
```Julia
julia> function main_qp()
           model = MOI.instantiate(
               UnoSolver.Optimizer;
               with_cache_type = Float64,
               with_bridge_type = Float64,
           )
           MOI.set(model, MOI.Silent(), true)
           x = MOI.add_variable(model)
           y = MOI.add_variable(model)
           c1 = MOI.add_constraint(model, 1.0 * y, MOI.EqualTo(1.0))
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           obj = 2.0 * x * x + 1.0 * y
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj)}(), obj)
           MOI.optimize!(model)
           MOI.get(model, MOI.ConstraintDual(), c1)
       end
main_qp (generic function with 1 method)

julia> function main_lp()
           model = MOI.instantiate(
               UnoSolver.Optimizer;
               with_cache_type = Float64,
               with_bridge_type = Float64,
           )
           MOI.set(model, MOI.Silent(), true)
           x = MOI.add_variable(model)
           y = MOI.add_variable(model)
           c1 = MOI.add_constraint(model, 1.0 * y, MOI.EqualTo(1.0))
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           obj = 1.0 * y
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj)}(), obj)
           MOI.optimize!(model)
           MOI.get(model, MOI.ConstraintDual(), c1)
       end
main_lp (generic function with 1 method)

julia> function main_qcp()
           model = MOI.instantiate(
               UnoSolver.Optimizer;
               with_cache_type = Float64,
               with_bridge_type = Float64,
           )
           MOI.set(model, MOI.Silent(), true)
           x = MOI.add_variable(model)
           y = MOI.add_variable(model)
           c1 = MOI.add_constraint(model, 1.0 * y, MOI.EqualTo(1.0))
           MOI.add_constraint(model, 1.0 * x * x, MOI.LessThan(1.0))
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           obj = 1.0 * y
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj)}(), obj)
           MOI.optimize!(model)
           MOI.get(model, MOI.ConstraintDual(), c1)
       end
main_qcp (generic function with 1 method)

julia> main_lp()
1.0

julia> main_qp()
-1.0

julia> main_qcp()
-1.0
```

interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
amontoison: In that case we probably need to modify the previous function
```julia
MOI.get(
    model::Optimizer,
    attr::MOI.ConstraintDual,
    ci::MOI.ConstraintIndex{<:_FUNCTIONS,<:_SETS},
)
```
because
```julia
const _FUNCTIONS = Union{
    MOI.ScalarAffineFunction{Float64},
    MOI.ScalarQuadraticFunction{Float64},
    MOI.ScalarNonlinearFunction,
}
```

odow:
Okay, no, it's that LP has a different dual convention to anything else...
```Julia

julia> function main_nlp()
           model = MOI.instantiate(
               UnoSolver.Optimizer;
               with_cache_type = Float64,
               with_bridge_type = Float64,
           )
           MOI.set(model, MOI.Silent(), true)
           x = MOI.add_variable(model)
           y = MOI.add_variable(model)
           c1 = MOI.add_constraint(model, 1.0 * y, MOI.EqualTo(1.0))
           f = MOI.ScalarNonlinearFunction(:exp, Any[x])
           MOI.add_constraint(model, f, MOI.LessThan(1.0))
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           obj = 1.0 * y
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj)}(), obj)
           MOI.optimize!(model)
           MOI.get(model, MOI.ConstraintDual(), c1)
       end
main_nlp (generic function with 1 method)

julia> main_nlp()
-1.0
```

interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
odow: Sure. But method specificity means it doesn't matter.

Anyway. I'm just trying to get the tests to pass for now.

amontoison:
> @amontoison there are still some failing tests that I need to debug.

Thanks a lot for your help Oscar!

odow:
@cvanaret have you documented the dual conventions anywhere? Why does adding a non-binding QCP constraint change the dual of a linear constraint?

odow:
I managed to get the tests to pass, but please don't merge. I think it's clearly wrong, and we just don't have enough tests.

I'd like to know the actual dual convention, and how/why it depends on the problem type and the various constraint types.

@cvanaret if it's not too late: strongly encourage you to consider the MOI convention. The dual value belongs the dual set. `f(x) >= a` implies that the dual is non-negative, `f(x) <= a` implies that the dual is negative. `f(x) = b` matches the binding direction. `f(x) in [l, u]` matches the binding side. The objective and problem types don't matter.

cvanaret:
Thanks for looking into it, Oscar! These dual conventions are a big mess :)

If I understand correctly, in MOI you stick to your dual convention (active lower bound -> positive dual and active upper bound -> negative dual) regardless of the optimization sense, while Uno flips this convention when maximizing. @leyffer  and I just had our weekly meeting and we agreed that it makes sense to have two dual conventions in Uno, one for minimization and one for maximization (you expect different signs for $\frac{\partial f}{\partial x_i}$ at active bounds, depending whether you're minimizing or maximizing).
The Lagrangian definition ($f(x) - y^T c(x) - z^T x$ or $f(x) + y^T c(x) + z^T x$) in Uno also has an influence on the dual signs (fortunately it's fixed for MOI). I need to take a moment to review all the possible cases.

The error you caught with the different dual convention for LPs is probably the special case for fixed variables (lower bound = upper bound) [that I implemented in BQPD](https://github.com/cvanaret/Uno/blob/main/uno/ingredients/subproblem_solvers/BQPD/BQPDSolver.cpp#L272). Apparently wrong as well :sweat_smile:

odow:
>  in MOI you stick to your dual convention (active lower bound -> positive dual and active upper bound -> negative dual) regardless of the optimization sense

Yes. The convention comes from conic duality. A lower bound is `f(x) - l in Nonnegatives()` and an upper bound is `f(x) - u in Nonpositives()`, and the cones are self dual, so dual of `f(x) >= l` must belong to `Nonnegatives()` and likewise for Nonpositives.

Once you depart from that, it becomes very messy.

odow:
Okay, so I've added some new comprehensive tests. 

We're currently failing a few tests where MOI thinks the problem is an NLP (because there are `MOI.ScalarNonlinearFunction`), but Uno can deduce that the underlying problem is actually linear (I assume, because the Hessian is 0).

If we keep the current duality convention (because changing it would be a breaking change to Uno), then we need to expose the `get_problem_type` at the C API.

odow:
Okay, I think I have everything working and passing tests.

From what I can tell, your dual convention is the following:

The dual of $f(x) \ge 0$ is of sign:

|     | min | max |
| --- | --- | --- |
| LP  | +ve | -ve |
| QP  | -ve | +ve |
| NLP | -ve | +ve |

LP follows the standard textbook definition of a shadow price (a one unit increase in the RHS gives a corresponding change in the objective of \pi). QP and NLP use the opposite (a one unit relaxation of the constraint gives a corresponding change in the objective of \pi).

amontoison:
Thanks Oscar for working on that.
Can you also uncomment the following lines: https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/test/MOI_wrapper.jl#L813-L970 ?

@cvanaret You should document somewhere what you do internally in Uno for the dual conventions.

interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1656,73 +1660,33 @@ function MOI.get(
...
-    nquad = length(model.qp_data)
-    v = Vector{Float64}(undef, model.inner.ncon - nquad)
-    for i = nquad+1:model.inner.ncon
-        v[i-nquad] = s * UnoSolver.uno_get_constraint_dual_solution_component(model.solver, i-1)
-    end
-    return v
+    s = _dual_multiplier(model)
+    return Float64[
+        s * UnoSolver.uno_get_constraint_dual_solution_component(model.solver, i)
+        for i in length(model.qp_data):(model.inner.ncon-1)
```
amontoison, line 1677: @odow If have non-linear contraints `ScalarNonlinearFunction` in the JuMP model and also a `MOI.VectorNonlinearOracle`.
Should we return all the dual solution of the nonlinear constraints or only the ones in the `NLPBlock`?
I may have done a mistake here.

odow: Only the ones in the NLPBlock. But really, people should stop using the NLPBlock. It's only for the legacy `@NL` macros

amontoison: And can we mix the legacy `@NL` with a oracle?
If the answer is yes, I will fix it after this PR.

odow: Nope

odow:
Fixed

cvanaret:
Thanks so much Oscar!
Uno only looks at `problem_type` (`LP`, `QP`, or `NLP`), since the Hessian may not be provided.


> From what I can tell, your dual convention is the following:
> 
> The dual of f ( x ) ≥ 0 is of sign:
> 	min 	max
> LP 	+ve 	-ve
> QP 	-ve 	+ve
> NLP 	-ve 	+ve

I wish I did that on purpose :expressionless:
I'll keep it like that for the moment, but having a clean convention like in conic optimization is quite attractive.

> @cvanaret You should document somewhere what you do internally in Uno for the dual conventions.

I will!

odow:
> Uno only looks at problem_type (LP, QP, or NLP)

This doesn't seem to be the case, since when I passed `problem_type = "NLP"` I got the LP dual convention if the problem was actually linear.

PR #402, opened 2025-10-29
[Julia] Test Uno with HiGHS v1.12

amontoison:
cc @odow


odow:
Yeah, but this requires you to recompile Uno_jll right?

odow:
It could be that HiGHS now includes OpenBLAS as a dependency.

amontoison:
I saw that you didn't compiled it with LBT but it should be ok (in theory).

amontoison:
Ok, we don't have issue with HiGHS v1.12.0, I can remove the compat entry on it.
The official `Uno_jll` compiled with Yggdrasil doesn't have a strict compat on `HiGHS_jll` like here so we are all good.

PR #403, opened 2025-10-29
[UnoSolver.jl] Add a generic function uno

amontoison:
@cvanaret Is it ok for you to add `SolverCore.jl` as a dependency?
https://github.com/JuliaSmoothOptimizers/SolverCore.jl

It is very light and I can have the same output than the other optimization solvers `madnlp`, `madncl`, `madipm`, `ipopt`, `knitro`, ....

@tmigot, Iam wondering if we should not re-export the symbols of `SolverCore.jl` in `NLPModels.jl` such that I only have`NLPModel.jl` for an extension and I don't need to rely on `SolverCore.jl` as a direct dependency.

### DIFF
--- a/interfaces/Julia/README.md
+++ b/interfaces/Julia/README.md
@@ -30,9 +30,7 @@ Below are examples showing how to use `UnoSolver.jl` with the interfaces for `NL
 using UnoSolver, CUTEst
 
 nlp = CUTEstModel{Float64}("HS15")
-model = uno_model(nlp)
-solver = uno_solver(preset="filtersqp", print_solution=true, logger="INFO")
-uno_optimize(solver, model)
+model, solver = uno(nlp, preset="filtersqp", print_solution=true, logger="INFO")
 
 timer = UnoSolver.uno_get_cpu_time(solver)
 niter = UnoSolver.uno_get_number_iterations(solver)


cvanaret:
@amontoison sure, works for me!

tmigot:
Thanks @cvanaret @amontoison for this, that's a great add to Julia community.

tmigot:
@amontoison you suggest that SolverCore becomes a dependency of NLPModels ?
I don't think it's possible from a design point of view. Both packages should be very light, if there is an issue with that please open an issue on NLPModels.jl.

amontoison:
> @amontoison you suggest that SolverCore becomes a dependency of NLPModels ? I don't think it's possible from a design point of view. Both packages should be very light, if there is an issue with that please open an issue on NLPModels.jl.

@tmigot I read the documentation of `SolverCore.jl`, and I now better understand the issue.
Unlike MOI, we have split the modeling and solver parts between `NLPModels.jl` and `SolverCore.jl`.
I asked this question because if I want to return a `GenericExecutionStats` in the `uno` function defined in the extension, I need to either condition the extension's installation on both packages (`NLPModels.jl` and `SolverCore.jl`), or add `SolverCore.jl` as a direct dependency of `UnoSolver.jl`.

I think adding `SolverCore.jl` as a direct dependency is the most relevant choice, since it still makes sense to use it even when the model is not an NLPModel.

Maybe something to share with the JSO team: https://github.com/cvanaret/Uno/tree/main/interfaces/Julia

interfaces/Julia/src/C_wrapper.jl
```diff
@@ -226,6 +226,43 @@ function uno_model(
...
+  hessian_triangle::Char='L',
+  lagrangian_sign::Float64=1.0;
+  kwargs...
+)
+  model = uno_model(problem_type, minimize, nvar, ncon, lvar, uvar, lcon, ucon, jrows,
+                    jcols, nnzj, hrows, hcols, nnzh, eval_objective, eval_constraints,
+                    eval_gradient, eval_jacobian, eval_hessian, eval_Jv, eval_Jtv,
+                    eval_Hv, user_model, hessian_triangle, lagrangian_sign)
+  solver = uno_solver(; kwargs...)
+  uno_optimize(solver, Model)
```
amontoison:
```suggestion
  uno_optimize(solver, model)
```

interfaces/Julia/ext/UnoSolverNLPModelsExt/UnoSolverNLPModelsExt.jl
```diff
@@ -75,14 +75,20 @@ function UnoSolver.uno_model(nlp::AbstractNLPModel{Float64}; operators_available
...
+    1.0,
   )
   UnoSolver.uno_set_initial_primal_iterate(model, nlp.meta.x0)
   UnoSolver.uno_set_initial_dual_iterate(model, nlp.meta.y0)
   return model
 end
 
+function UnoSolver.uno(nlp::AbstractNLPModel{Float64, Vector{Float64}}, operators_available::Bool=true; kwargs...)
+  model = UnoSolver.uno_model(nlp, operators_available)
+  solver = UnoSolver.uno_solver(; kwargs...)
```
amontoison, line 90:
```suggestion
  solver = UnoSolver.uno_solver(; kwargs...)
  uno_optimize(solver, model)
```

amontoison:
Ok, I like the current modifications.
I just want to change the output of the new generic function `uno` to return an `AbstractExecutionStats` from `SolverCore.jl` and it should be good.
I may do it in a different PR.

PR #407, opened 2025-11-03
[CI] Upgrade macos-13 to macos-latest

cvanaret:
The macOS 13 runner image will be retired by December 4th, 2025. This deprecation includes the following labels:
- macos-13
- macos-13-large
- macos-13-xlarge

### DIFF
--- a/.github/workflows/julia-tests-jll.yml
+++ b/.github/workflows/julia-tests-jll.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         version: ['1']
-        os: [ubuntu-latest, windows-latest, macos-13]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         arch: ['x64']
         allow_failure: [false]
         include:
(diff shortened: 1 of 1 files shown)


amontoison (approved):

imciner2:
The `macos-latest` runner image is an aarch64 runner image, but it looks like it is in the `x64` architecture block here. This means you will be requesting the x64 Julia version on the arm64 machine, so it goes through the emulation layer - this then triggers the new warning added in setup-julia: https://github.com/julia-actions/setup-julia/pull/300. There was a discussion on the Julia Slack last month actually about running the x64 version on arm64 in CI having problems with artifacts loading properly. I would suggest that if you want an x64 architecture for macOS here, use the `macos-15-intel` runner image for that.

amontoison:
I tried it and I can't install the Fortran compiler for `CUTEst.jl`. I don't know why the system is unable to find `gfortran`. I just changed `macos-13` into `macos15-intel`.

PR #426, opened 2025-11-13
Add function in interfaces to get the solver's method

### LINKED ISSUE #421: Add function in interfaces to get the algorithm description
Uno generates a description of the strategy combination like:
`TR Fletcher-filter restoration inequality-constrained method with exact Hessian and no regularization`
or
`LS Waechter-filter restoration primal-dual interior-point method with exact Hessian and primal-dual regularization`

I think it would be beneficial to have a function in the interfaces that returns this description, such that the user can see the algorithm that is used.

cc @amontoison

cvanaret:
Added a function `uno_get_solver_method()` in C/Fortran/Python interfaces for the user to check which method is used by the solver, based on the strategy combination.
This is especially useful because some special cases in the ingredient instantiation exist (e.g., if unconstrained, Uno automatically picks merit function + no relaxation).

Fixes https://github.com/cvanaret/Uno/issues/421

### DIFF
--- a/interfaces/C/Uno_C_API.cpp
+++ b/interfaces/C/Uno_C_API.cpp
@@ -834,6 +834,12 @@ Result* uno_get_result(void* solver) {
    return uno_solver->result;
 }
 
+const char* uno_get_solver_method(void* solver) {
+   assert(solver != nullptr);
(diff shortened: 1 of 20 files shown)


cvanaret:
Superseded by https://github.com/cvanaret/Uno/pull/719

PR #434, opened 2025-11-17
[UnoSolver.jl] parse Bool options properly

### LINKED ISSUE #433: [UnoSolver.jl] Bool options are not properly parsed
@amontoison this one is nasty 😆 
`Bool` options are not parsed properly: in Julia `Bool` is a subset of `Integer`, so `Bool` options are parsed as `Integer`...
See [here](https://github.com/cvanaret/Uno/blob/81262ccb4fe5a16e6c536eecba49067cfe42dbe1/interfaces/Julia/src/C_wrapper.jl#L322).

cvanaret:
Fixes #433


amontoison:
@cvanaret You also need to do the same modification in the extension of MOI.
See the file `MOI_wrapper.jl`.

cvanaret:
> @cvanaret You also need to do the same modification in the extension of MOI. See the file `MOI_wrapper.jl`.

Thanks :slightly_smiling_face:

PR #444, opened 2025-11-18
[UnoSolver.jl] Use more optimized hess_coord! and hprod! for unconstrained problems

### LINKED ISSUE #443: Segfaults or mysterious `trunc` crashes with `UnoSolver.jl` in a custom `NLPModel`
Hi @cvanaret 

First, thanks so much again for creating such an exciting new solver and releasing it as libre software. I'm very deeply grateful. 

TL;DR: I have a custom `NLPModel` with only box constraints that I can hand to `NLPModelsIpopt.ipopt` that works fine, but does not work when I hand it to `UnoSolver` via `uno(nlp, false; preset="filtersqp")` because I get a method error about not implementing `hess_coord!` in the constrained case. Considering that that method shouldn't (?) be called in a purely box-constrained case, I think there is some kind of communication issue between my `NLPModel`-type object and `uno`, but I can't quite figure out what it is.

Longer story:

I am trying to incorporate `UnoSolver.jl` as the de facto standard optimizer in my Gaussian process model fitting package [`Vecchia.jl`](https://github.com/cgeoga/Vecchia.jl) using the extension system. At the moment, the solution that I am trying is to write an extension for `NLPModels.jl`, implement my `<: AbstractNLPModel` struct with the appropriate methods, and then have users pass in the necessary fields. This object of mine does _not_ implement `NLPModels.hess_prod!` with the extra argument for the case of having constraints, although I have added a dummy version for now as I try to track down the issue here.

In the process of implementing the `UnoSolver.jl` bindings, however, I'm hitting some funny segfaults or mysterious method errors. To start, the following code snippet works:
```julia
# full file of Vecchia.jl/example/example_estimate_ipopt.jl (comments removed)
using ForwardDiff, NLPModels, NLPModelsIpopt

include("example_setup.jl")
const cfg = Vecchia.knnconfig(sim, pts, 10, matern)
solver    = NLPModelsSolver(ipopt, Dict(:tol=>1e-4))
estimator = vecchia_estimate(cfg, init[1:3], solver) 
```
And `vecchia_estimate` internally calls `Vecchia.optimize` in the `VecchiaNLPModelsExt` code [here](https://github.com/cgeoga/Vecchia.jl/blob/main/ext/VecchiaNLPModelsExt.jl#L80). So I think my `NLPModel` object is not (completely) inappropriately designed. 

My problems are low-dimensional with very expensive gradients and Hessians, and so I wish to disallow using Hvps. After chatting with @amontoison I understand that I need to communicate that as follows (code link [here](https://github.com/cgeoga/Vecchia.jl/blob/57daedd49ec46e46364bc4f361bdf4acd44351e3/ext/VecchiaUnoSolverExt.jl#L11)):
```julia
    # snippet from Vecchia.jl/ext/VecchiaUnoSolverExt.jl
    nlp = Vecchia.nlp(obj, init, box_lower=box_lower, box_upper=box_upper)
    (model, solver) = uno(nlp, false; preset="filtersqp", 
                          print_solution=false, operator_available=false)
    UnoSolver.uno_get_primal_solution(solver, zeros(length(init)))
  end
```
But here is the rub: when I run this example code with
```julia
# full file of Vecchia.jl/example_estimate_uno.jl
using NLPModels, ForwardDiff, UnoSolver

include("example_setup.jl")
const cfg = knnconfig(sim, pts, 10, matern)

# this code triggers the function I'm showing a snippet of from Vecchia.jl/ext/VecchiaUnoSolverExt.jl
est = Vecchia.optimize(cfg, init[1:3], Vecchia.UnoNLPSolver();
                       box_lower=[1e-8, 1e-8, 0.25], 
                       box_upper=[10.0, 10.0, 5.0])
```
I crash with a segfault or get an error of `trunc(UInt64, -138[...])` in `uno_objective_gradient`. However, if I add an early return to the snippet from the `VecchiaUnoSolverExt.jl` file linked above and then _separately_ hand that to `uno` with
```julia
(model, solver) = uno(est, false; preset="filtersqp", print_solution=false)
```
Everything runs great so long as I have the dummy method for `hess_prod!` that takes the extra variable but does nothing with it. 

I have a theory about what's going on, but I would not say I am very confident. I think that maybe the way that `NLPModels.jl` is communicating about box constraints with `Uno` is not quite working right. In my case where I am just box constrained, for example, I don't think I should have to implement `NLPModels.hess_coord!` with the extra argument `y` for constraints. But if I don't do that, even in the manual process of returning `nlp` and then handing it to `uno` throws a method error of 
```julia
ERROR: MethodError: no method matching hess_coord!(::VecchiaNLPModelsExt.VecchiaNLPModel{…}, ::Vector{…}, ::Vector{…}, ::Vector{…}; obj_weight::Float64)
```

I would appreciate any thoughts or suggestions! Thank you again for the wonderful software.

amontoison:
close #443
cc @cgeoga

### DIFF
--- a/interfaces/Julia/ext/UnoSolverNLPModelsExt/UnoSolverNLPModelsExt.jl
+++ b/interfaces/Julia/ext/UnoSolverNLPModelsExt/UnoSolverNLPModelsExt.jl
@@ -26,7 +26,11 @@ end
 
 function nlpmodels_lagrangian_hessian(nlp::AbstractNLPModel{Float64, Vector{Float64}}, hvals::Vector{Float64}, x::Vector{Float64},
                                       multipliers::Vector{Float64}, objective_multiplier::Float64)
-NLPModels.hess_coord!(nlp, x, multipliers, hvals, obj_weight=objective_multiplier)
+  if nlp.meta.ncon == 0
+    NLPModels.hess_coord!(nlp, x, hvals, obj_weight=objective_multiplier)
+  else
+    NLPModels.hess_coord!(nlp, x, multipliers, hvals, obj_weight=objective_multiplier)
+  end
   return hvals
 end
 
@@ -45,7 +49,11 @@ end
 function nlpmodels_lagrangian_hessian_operator(nlp::AbstractNLPModel{Float64, Vector{Float64}}, Hv::Vector{Float64},
                                                x::Vector{Float64}, objective_multiplier::Float64, multipliers::Vector{Float64},
                                                v::Vector{Float64}, evaluate_at_x::Bool)
-  NLPModels.hprod!(nlp, x, multipliers, v, Hv; obj_weight=objective_multiplier)
+  if nlp.meta.ncon == 0
+    NLPModels.hprod!(nlp, x, v, Hv; obj_weight=objective_multiplier)
+  else
+    NLPModels.hprod!(nlp, x, multipliers, v, Hv; obj_weight=objective_multiplier)
+  end
   return Hv
 end
 


interfaces/Julia/ext/UnoSolverNLPModelsExt/UnoSolverNLPModelsExt.jl
```diff
@@ -26,7 +26,11 @@ end
 
 function nlpmodels_lagrangian_hessian(nlp::AbstractNLPModel{Float64, Vector{Float64}}, hvals::Vector{Float64}, x::Vector{Float64},
```
cvanaret, line 27: Would that be an option to overload `nlpmodels_lagrangian_hessian` with no `multipliers` parameter? Otherwise the user still has to provide them (empty).

amontoison: ?

amontoison: `nlp.meta.ncon` stores the number of constraints in the model.
And we have an API in `NLPModel.jl` for Hessian / H*v of the objective and Hessian / H_L * v of the Lagrangian.

I just call the API for unconstrained problems in we don't have constrained.

cvanaret: I meant to have both:
```julia
function nlpmodels_lagrangian_hessian(nlp::AbstractNLPModel{Float64, Vector{Float64}}, hvals::Vector{Float64}, x::Vector{Float64},
                                      multipliers::Vector{Float64}, objective_multiplier::Float64)
  NLPModels.hess_coord!(nlp, x, multipliers, hvals, obj_weight=objective_multiplier)
  return hvals
end
```
and
```julia
function nlpmodels_lagrangian_hessian(nlp::AbstractNLPModel{Float64, Vector{Float64}}, hvals::Vector{Float64}, x::Vector{Float64},
                                      objective_multiplier::Float64)
  NLPModels.hess_coord!(nlp, x, hvals, obj_weight=objective_multiplier)
  return hvals
end
```
so that Chris can call the second `nlpmodels_lagrangian_hessian` directly.

amontoison: > Would that be an option to overload `nlpmodels_lagrangian_hessian` with no `multipliers` parameter? Otherwise the user still has to provide them (empty).

Oh, I think I understand what you wanted to say. This function is just the C function passed to Uno. I don't want to change the C interface.

But inside this "C function", I can call the best Julia funtion depending if the problem is constrained or not.

cvanaret: Ah I see! Yes, if it's passed to Uno, we have to stick to the one signature.

amontoison: > I meant to have both:
> 
> ```julia
> function nlpmodels_lagrangian_hessian(nlp::AbstractNLPModel{Float64, Vector{Float64}}, hvals::Vector{Float64}, x::Vector{Float64},
>                                       multipliers::Vector{Float64}, objective_multiplier::Float64)
>   NLPModels.hess_coord!(nlp, x, multipliers, hvals, obj_weight=objective_multiplier)
>   return hvals
> end
> ```
> 
> and
> 
> ```julia
> function nlpmodels_lagrangian_hessian(nlp::AbstractNLPModel{Float64, Vector{Float64}}, hvals::Vector{Float64}, x::Vector{Float64},
>                                       objective_multiplier::Float64)
>   NLPModels.hess_coord!(nlp, x, hvals, obj_weight=objective_multiplier)
>   return hvals
> end
> ```
> 
> so that Chris can call the second `nlpmodels_lagrangian_hessian` directly.

We can do that but I don't want to provide two API for the Hessian in the C interface if the problem is unconstrained or not.

It is best thing to do if we want more bugs!!

It is better to handle it on the Julia side.

cvanaret: Agreed! Your version looks good.

cvanaret (approved):

PR #447, opened 2025-11-20
[UnoSolver.jl] Release 0.1.5

cvanaret:


amontoison:
It is not a breaking release. Keep it as a patch.
Before the version 1.0, minor release == breaking release.

interfaces/Julia/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "UnoSolver"
 uuid = "1baa60ac-02f7-4b39-a7a8-2f4f58486b05"
 authors = ["Alexis Montoison <amontoison@anl.gov>", "Charlie Vanaret <vanaret@zib.de>"]
-version = "0.1.4"
+version = "0.2.0"
```
amontoison: 9
```suggestion
version = "0.1.5"
```

cvanaret:
Ah, we now see the exact same issue with an `NLPModel` test: `DimensionError: Input x should have length 2 not 1824537376`.
Most likely the typos prevented us from seeing it :sweat_smile:

amontoison:
Update: I am tired, it is all good on the Julia side.

cvanaret:
No worries, I'll investigate :) Looks like the call to the Jacobian evaluation isn't working.

cvanaret:
@amontoison I would release the patch anyway, if that's ok.

amontoison:
Yep

PR #452, opened 2025-11-25
[UnoSolver.jl] Clean up MOI interface

cvanaret:
Replaced `model` with `optimizer` and `inner` with `model`.


amontoison:
@cvanaret I like to follow the same variable names than `Ipopt.jl`.
If they accept the same modifications in `Ipopt.jl`, I agree to merge this PR.
It is way easier to maintain for me this way...

cvanaret:
@amontoison got it, happy to close the PR. This is just a very confusing notation.

amontoison:
@odow What do you think of the modifications of Charlie? I am ok to use the same one in Ipopt / Uno / MadNLP.

odow:
I'm not making changes to Ipopt at this point. Sure, `model.inner` is not part of the public API, but I'm sure plenty of people still call it.

PR #455, opened 2025-11-28
Detect NaNs early

arnavk23:
- I made `clone()` and `contains_nan()` optional defaults in `EvaluationSpace`. This keeps compatibility with many existing `EvaluationSpace` subclasses; only subclasses that we explicitly enhanced will clone/inspect their stored data. If `clone()` returns nullptr, the early NaN check is skipped for that evaluation space (but exceptions from evaluating trial functions will still be caught and cause rejection).
- I also implemented `clone()` and `contains_nan()`  for the three major evaluation-space types used by direct solvers and QP (COO, BQPD, HiGHS). If you want the same behavior (clone + NaN detection) for other evaluation-space implementations, I can add the overrides the same way.
- I implemented detection (reject trial iterate) but did not implement swapping of solver internal evaluation spaces. That requires adding trial/current spaces inside solver classes (or another coordinated approach) across all solvers — I can implement that next if you want. The current implementation prevents NaN trial iterates from being accepted, which addresses the "detect NaNs early" part; swapping to reuse trial evaluations is a follow-up.

### DIFF
--- a/uno/ingredients/subproblem_solvers/COOEvaluationSpace.cpp
+++ b/uno/ingredients/subproblem_solvers/COOEvaluationSpace.cpp
@@ -1,7 +1,8 @@
 // Copyright (c) 2025 Charlie Vanaret
-// Licensed under the MIT license. See LICENSE file in the project directory for
-// details.
+// Licensed under the MIT license. See LICENSE file in the project directory for details.
 
+#include <algorithm>
+#include <stdexcept>
 #include "COOEvaluationSpace.hpp"
 #include "ingredients/subproblem/Subproblem.hpp"
 #include "ingredients/subproblem_solvers/DirectSymmetricIndefiniteLinearSolver.hpp"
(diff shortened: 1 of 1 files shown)


cvanaret:
Hi @arnavk23, thanks for looking into this! I realized that we shouldn't have two copies of `EvaluationSpace` because we'll have redundant sparsity patterns and so on. So we want both current and trial evaluations in the same object.
It's actually more complex than that (we also want to separate model and problem evaluations). It's closely connected to the L-BFGS developments I'm working on, so I'll deal with it :wink:

arnavk23:
> Hi @arnavk23, thanks for looking into this! I realized that we shouldn't have two copies of `EvaluationSpace` because we'll have redundant sparsity patterns and so on. So we want both current and trial evaluations in the same object.
> It's actually more complex than that (we also want to separate model and problem evaluations). It's closely connected to the L-BFGS developments I'm working on, so I'll deal with it :wink: 

Great to hear. Should I close the pr then

cvanaret:
You can leave it as is, I'll close it when mine is merged.

PR #457, opened 2025-12-01
[UnoSolver.jl] Improve the fix with GC

amontoison:
@cvanaret Can you confirm that you do all asserts in the C++ code?
It is a better fix for #443.

### DIFF
--- a/interfaces/C/Uno_C_API.cpp
+++ b/interfaces/C/Uno_C_API.cpp
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project directory for details.
 
 #include <algorithm>
-#include <cassert>
 #include <cstring>
 #include <iostream>
 #include <streambuf>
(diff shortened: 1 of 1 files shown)


interfaces/Julia/src/C_wrapper.jl
```diff
@@ -275,24 +273,24 @@ function uno(
...
-  end
+  uno_optimize(solver, model)
   return model, solver
 end
 
 function uno_set_initial_primal_iterate(model::Model, initial_primal_iterate::Vector{Float64})
-  @assert model.c_model != C_NULL
   @assert model.nvar == length(initial_primal_iterate)
-  flag = uno_set_initial_primal_iterate(model.c_model, initial_primal_iterate)
+  GC.@preserve model begin
```
cvanaret, line 282: If I understand correctly, you need to `GC.preserve` only when you pass `model.c_model` or `solver.c_solver` to a function call?

cvanaret:
> @cvanaret Can you confirm that you do all asserts in the C++ code? It is a better fix for #443.

I have asserts in the C API, but they're tested only in `Debug` mode. Nothing is checked in `Release` mode.
For setter functions, I could return a boolean (false upon failure), but for getter functions, it would be more involved (like returning a boolean, and having a pointer as argument to store the result upon success).

cvanaret:
Ended up:
- returning `false` in setters,
- throwing an exception in getters,

if pointers to model or solver are invalid in `Uno_C_API.cpp`.

amontoison:
Thanks Charlie!

Issue #374, opened 2025-10-22
[UnoSolver.jl] Add an example with the HSL linear solvers


amontoison:
It will be great to explain how we can use `MA27` or `MA57` in `UnoSolver.jl`.
I did it for `Ipopt.jl` a few years ago: https://github.com/jump-dev/Ipopt.jl?tab=readme-ov-file#linear-solvers

cvanaret:
@amontoison I can write a draft and ask you for a review.

amontoison:
@cvanaret It is fine for me for if you have the bandwidth!

amontoison:
@cvanaret bump

PR #460, opened 2025-12-02
[UnoSolver.jl] how to use subproblem solvers in README

cvanaret:
Fixes https://github.com/cvanaret/Uno/issues/374

### DIFF
--- a/interfaces/Julia/README.md
+++ b/interfaces/Julia/README.md
@@ -102,7 +102,7 @@ uno_optimize(solver, model)
 
 ## Linear solvers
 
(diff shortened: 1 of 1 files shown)


amontoison:
We should also specify that HSL solvers are used by default in Uno if they are available.
It is not the case for Ipopt (except if the official `libhsl.a` is available at compilation time.

cvanaret:
> We should also specify that HSL solvers are used by default in Uno if they are available. It is not the case for Ipopt (except if the official `libhsl.a` is available at compilation time.

Good idea!

amontoison (approved):

PR #461, opened 2025-12-03
CMake: automatically download the BQPD artifact and link it

cvanaret:
There's apparently no easy way to link the BQPD static library during the build of `unopy`, since a modern packaging system expects a plain `pyproject.toml` such as:
```
[build-system]
requires = ["scikit-build-core", "pybind11"]
build-backend = "scikit_build_core.build"

[project]
name = "example"
version = "0.1.0"
```

To fetch the BQPD artifact, we would need an actual script (`setup.py` like in https://github.com/cvanaret/Uno/pull/311) but this is not the recommended way to do it.

Here I'm trying to fetch and link the BQPD artifact directly in CMake, which means that Uno, `UnoSolver.jl` and `unopy` would be systematically built with BQPD. The other dependencies can be either statically linked or detected at runtime via dynamic loading.

Roadmap:
- [x] fetch the correct BQPD artifact
- [ ] link the BQPD static library
  - [x] Linux and macOS
  - [ ] Windows
- [ ] remove BQPD from the Github Actions
- [ ] remove BQPD from the `UnoSolver.jl` and `Uno_jll` dependencies

@amontoison does that sound like a good idea?


amontoison:
I am not a big fan of changing the current compilation in Julia.

I also don't understand the issue,  when `unopy` is compiled, `uno` is also compiled, right?
So if you also download BQPD, Cmake should autodetect it and compile Uno with it.

cvanaret:
IIUC, I can't do what I'm doing here (in the CMake) in the Python build. The `pyproject.toml` expects as dependencies either Python packages (which I cannot build for BQPD because of the missing symbol) or [Github source repos](https://chrisholdgraf.com/blog/2022/install-github-from-pyproject/). Github artifacts/plain static libraries are not something it can work with.

cvanaret:
Mh wait a minute. It looks like it could be possible to build the wheels and to upload them to PyPi in CI with [cibuildwheel](https://cibuildwheel.pypa.io/en/1.x/deliver-to-pypi/) :open_mouth: 
Perhaps there's a way to inject the dependencies at this point. On my way to ask Jari :smile:

PR #465, opened 2025-12-09
Use the new features of NLPModels.jl v0.21.6

amontoison:
It requires https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/525 before and a new release of `NLPModels.jl`.

If  Hessian-vector products are not available, we can easily get this information for the model now and avoid to use BQPD.

It should also help for the support of LBFGS (#150) by making the Hessian of the Lagrangian optional.

### DIFF
--- a/interfaces/Julia/test/MOI_wrapper.jl
+++ b/interfaces/Julia/test/MOI_wrapper.jl
@@ -54,7 +54,8 @@ function test_MOI_Test()
             r"^test_quadratic_SecondOrderCone_basic$",
             r"^test_quadratic_nonconvex_constraint_basic$",
             # These tests require the support of MOI.LagrangeMultiplier
-            r"^test_VectorNonlinearOracle_LagrangeMultipliers$",
+            r"^test_VectorNonlinearOracle_LagrangeMultipliers_MIN_SENSE$",
+            r"^test_VectorNonlinearOracle_LagrangeMultipliers_MAX_SENSE$",
         ],
     )
     return


amontoison:
@odow It seems that you broke the unit tests of `UnoSolver.jl` with the last release of `MOI`.
It is related to new tests with the Lagrange multipliers of the `VectorNonlinearOracle`.

Update: I will not run these tests and add the support for it later.

cvanaret:
@amontoison I think the Github servers had an issue, a lot of scripts couldn't download the dependencies. I restarted them.

amontoison:
Yes but it is only for the compilation tests. The unit tests of `UnoSolver.jl` failed because of new unit tests in MOI 1.48.

cvanaret:
Ah I see!

odow:
Yeah the "is modifying a test breaking" is a bit of an open question. I tend to go with "it isn't" because it doesn't change any user-facing behaviour. It's useful for you to know that there are new features in MOI to implement. I guess I could have added a new separate test, rather than modifying an existing one...

amontoison:
I agree with you that as long as we only break the tests, it doesn't impact the users and only a limited subset of developers.

What could be nice is to attach an attribute `VersionNumber` with each test (or new test).
For example, this `VersionNumber` can be the latest release of `MOI` before the test was added.

This way we could specify the version of MOI used to interface our package (like `v1.4.6` here) in `MOI.Test.runtests` (optional or keyword argument) and potentially only run the MOI tests added up to this version.
If a version number is not provided in `MOI.Test.runtests`, we still run all the tests like now (not breaking).

It is a suggestion 🙂

odow:
Already implemented: `MOI.Test.runtests(...; exclude_tests_after = "v1.46.0")`

It wouldn't have helped in this cause because I modified an existing test... I guess I would have also bumped the version.

PR #486, opened 2026-01-15
[Julia] Add a function uno_statistics

### LINKED ISSUE #405: [UnoSolver.jl] Return an AbstractExecutionStats for the Julia function uno
We will need to add [SolverCore.jl](https://github.com/JuliaSmoothOptimizers/SolverCore.jl) as a dependency.

amontoison:
Partially fix #405

- Add a structure `Statistics` in the Julia interface `UnoSolver.jl`
- The function `uno` now returns a structure `Statistics` and not anymore a tuple `(Model, Solver)`.
- Remove `ADNLPModels.jl` from the tests because it is a heavy test dependency and we fixed the issue with GC.

### DIFF
--- a/.github/workflows/julia-tests-uno.yml
+++ b/.github/workflows/julia-tests-uno.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           version: "1.7"
           arch: x64
-      # - uses: julia-actions/cache@v2
+      - uses: julia-actions/cache@v2
       # Set environment variables required by BinaryBuilder.
       - name: Set the environment variables
         run: |


amontoison:
@cvanaret bump

.github/workflows/julia-tests-uno.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           version: "1.7"
           arch: x64
-      - uses: julia-actions/cache@v2
+      # - uses: julia-actions/cache@v2
```
amontoison:
```suggestion
      - uses: julia-actions/cache@v2
```

interfaces/Julia/src/C_wrapper.jl
```diff
@@ -302,3 +262,102 @@ function uno_solver(; kwargs...)
...
+  number_hessian_evaluations = UnoSolver.uno_get_number_hessian_evaluations(solver)
+
+  stats = Statistics(model, solver, primal_solution, constraint_dual_solution, lower_bound_dual_solution, upper_bound_dual_solution,
+                     cpu_time, solution_objective, solution_primal_feasibility, solution_stationarity, solution_complementarity,
+                     optimization_status, solution_status, number_iterations, number_subproblem_solved_evaluations,
+                     number_objective_evaluations, number_constraint_evaluations, number_objective_gradient_evaluations,
+                     number_jacobian_evaluations, number_hessian_evaluations)
+end
+
+function uno(
```
cvanaret, line 325: Could we imagine renaming the `uno` function into `uno_optimize`?

amontoison: No, because `uno_optimize` is the C routine that update the C model.
I don't want to mix the C interface with the Julia interface.
Users could be also interested to only ask for a few statistics and not all of them.

PR #488, opened 2026-01-25
A scipy.optimize.minimize interface

FrancoisGallard:
This implements an interface that is compatible with the https://docs.scipy.org/doc/scipy-1.16.2/reference/generated/scipy.optimize.minimize.html method.

It is used a lot in Python, and it will be faster for many users to use this interface.

Currently it does not support the Hessian and Hvp. I tried to implement it in a different branch but it does not work yet (I dont know why).


cvanaret:
Thanks a lot @FrancoisGallard! I'll have a look at the submission. Anything I can help with regarding 2nd-order information?
It would also be nice to have a Github Action that runs `test_scipy_interface.py`.

FrancoisGallard:
> Thanks a lot @FrancoisGallard! I'll have a look at the submission. Anything I can help with regarding 2nd-order information? It would also be nice to have a Github Action that runs `test_scipy_interface.py`.

Sure for the github action. It seems mandatory. I dont know how to do that. Do you?

For the hessian, yes I think we could take some time to debug together but before maybe you can have a look at this branch there is a test for the hessian that fails. Its really weird it crashed when accessing the x value, but I think the real bug source is elsewhere, peobably the way I pass the hessian.

https://github.com/FrancoisGallard/Uno/tree/scipy_hess

cvanaret:
@FrancoisGallard gentle bump :) Let me know if I can help.

FrancoisGallard:
I just created a new PR with Hessian, its a re implementation with Claude and @jgiret https://github.com/cvanaret/Uno/pull/705

PR #497, opened 2026-02-03
Evaluate functions at current and trial iterates in centralized EvaluationCache

cvanaret:
- removed `Evaluations` from `Iterate`
- created `EvaluationCache` that contains Jacobian sparsity + current and trial `Evaluations`
- `EvaluationCache` and its members `current_evaluations`/`trial_evaluations` is passed around
- in `Evaluations`, the 0th and 1st order function evaluations are performed lazily, i.e. evaluated once (flags are maintained)
- when the trial iterate is rejected, the flags are reset (= functions must be re-evaluated)
- when the trial iterate is accepted,  `current_evaluations` and `trial_evaluations` are swapped so that the `trial_evaluations` can now be used as current evaluations


cvanaret:
This one makes me very happy.

cvanaret:
Note: because of a merge, the text `[unopy] Release` is present in the final commit message, which wrongly triggered a push to PyPi (rejected anyway). It's fine.

PR #498, opened 2026-02-04
[Fortran] Update the interface to better handle C-strings

amontoison:
We need to discuss about a convention with Charlie for the suffix of the routines / include files of the Fortran interface before that we merge this PR.

I am wondering if we should not use a suffix if we don't have ambiguities and `_c` / `_f` if it is problematic.
We can also have two APIs, with `_c` and `_f`, but we duplicate a lot of code...

For the headers, should we have `uno_c.f90` and `uno_f.f90` ?
I kept `uno.f90` for the original C wrappers but I am fine if we rename it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,7 +247,7 @@ if(BUILD_SHARED_LIBS)
    target_include_directories(uno_shared PUBLIC ${DIRECTORIES})
    target_link_libraries(uno_shared PUBLIC ${LIBRARIES} ${FORTRAN_LIBS})
    set_target_properties(uno_shared PROPERTIES OUTPUT_NAME "uno")
-   set_target_properties(uno_shared PROPERTIES PUBLIC_HEADER "interfaces/C/Uno_C_API.h;interfaces/C/uno_int.h;;interfaces/Fortran/uno_c.f90;interfaces/Fortran/uno_fortran.f90")
+   set_target_properties(uno_shared PROPERTIES PUBLIC_HEADER "interfaces/C/Uno_C_API.h;interfaces/C/uno_int.h;interfaces/Fortran/uno_c.f90;interfaces/Fortran/uno_fortran.f90")
 endif()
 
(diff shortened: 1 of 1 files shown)


amontoison:
@cvanaret If all tests passed, I merge the PR.

cvanaret:
@amontoison works for me!

CMakeLists.txt
```diff
@@ -239,15 +239,15 @@ if(BUILD_STATIC_LIBS)
...
+   set_target_properties(uno_static PROPERTIES PUBLIC_HEADER "interfaces/C/Uno_C_API.h;interfaces/C/uno_int.h;interfaces/Fortran/uno_c.f90;interfaces/Fortran/uno_fortran.f90")
 endif()
 
 if(BUILD_SHARED_LIBS)
    add_library(uno_shared SHARED $<TARGET_OBJECTS:compiled_uno_files>)
    target_include_directories(uno_shared PUBLIC ${DIRECTORIES})
    target_link_libraries(uno_shared PUBLIC ${LIBRARIES} ${FORTRAN_LIBS})
    set_target_properties(uno_shared PROPERTIES OUTPUT_NAME "uno")
-   set_target_properties(uno_shared PROPERTIES PUBLIC_HEADER "interfaces/C/Uno_C_API.h;interfaces/C/uno_int.h;interfaces/Fortran/uno.f90")
+   set_target_properties(uno_shared PROPERTIES PUBLIC_HEADER "interfaces/C/Uno_C_API.h;interfaces/C/uno_int.h;;interfaces/Fortran/uno_c.f90;interfaces/Fortran/uno_fortran.f90")
```
amontoison:
```suggestion
   set_target_properties(uno_shared PROPERTIES PUBLIC_HEADER "interfaces/C/Uno_C_API.h;interfaces/C/uno_int.h;interfaces/Fortran/uno_c.f90;interfaces/Fortran/uno_fortran.f90")
```

PR #500, opened 2026-02-05
[CI] Use HiGHS_jll v1.12.0

amontoison:


amontoison:
@odow We have some tests failing with the release 1.13 of HiGHS, but only with the AMPL interface:
https://github.com/cvanaret/Uno/actions/runs/21697262051/job/62575415462

The concerned code is in [AmplNLWriter.jl](https://github.com/jump-dev/AmplNLWriter.jl).

I do a quick fix for now.
The tests of `UnoSolver.jl` are not impacted.

odow:
`free(): invalid pointer` this doesn't seem like it's our problem?

amontoison:
> `free(): invalid pointer` this doesn't seem like it's our problem?

I didn't see this line. I will try to confirm that the issue is only in HiGHS 1.13.
It may just be a bug in HiGHS. :crossed_fingers:

.github/workflows/julia-tests-ubuntu.yml
```diff
@@ -63,4 +63,5 @@ jobs:
           using Pkg
           Pkg.instantiate()
           Pkg.develop(path="/home/runner/.julia/dev/Uno_jll")
+          Pkg.add(name="HiGHS_jll", version="1.12.0")
```
amontoison, line 66:
```suggestion
          Pkg.add(name="HiGHS_jll", version="1.12.0")
          Pkg.pin(name="HiGHS_jll", version="1.12.0")
```

odow:
Yeah this is either a bug in HiGHS or a bug in how you're compiling Uno. I don't think it's anything to do with JuMP/Julia/MOI.

amontoison:
I agree with you. It is working with HiGHS 1.12 so I highly expect a bug in HiGHS 1.13, except if they did a small breaking change in the code and we now need to update something in Uno.

Issue #209, opened 2025-07-05
[CI] Test Uno with HiGHS


amontoison:
It will be great to test Uno with HiGHS.
We can do something to what was done for MUMPS (#190).

PR #509, opened 2026-02-10
[CI] Compile Uno with HiGHS

cvanaret:
- Linked HiGHS in addition to BQPD and MUMPS
- Used CMake's `FetchContent` to download `GTest`. Changed unit tests workflows accordingly

Fixes https://github.com/cvanaret/Uno/issues/209

### DIFF
--- a/.github/julia/build_tarballs_release.jl
+++ b/.github/julia/build_tarballs_release.jl
@@ -87,12 +87,6 @@ cd $WORKSPACE/srcdir/Uno
 mkdir -p build
 cd build
 
-if [[ "${target}" == *mingw* ]]; then
-    HIGHS_DIR=${prefix}/lib
-else
-    HIGHS_DIR=${libdir}
-fi
-
 if [[ "${target}" == *apple* ]] || [[ "${target}" == *freebsd* ]]; then
     OMP=omp
 else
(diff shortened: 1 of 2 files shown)


cvanaret:
OK, using CMake's `FetchContent`, I was able to link `googletest` on Windows/MinGW. Also adapted the other unit tests workflows.

amontoison:
@cvanaret Can you also update `build_tarballs release.jl` and `build_tarballs_yggsrasil.jl` to use the new option `-DHIGHS`?

cvanaret:
> @cvanaret Can you also update `build_tarballs release.jl` and `build_tarballs_yggsrasil.jl` to use the new option `-DHIGHS`?

I'm on it!

cvanaret:
@amontoison now the BinaryBuilder builds on Windows fail with:
```
No rule to make target '/workspace/destdir/lib/libhighs.dll', needed by 'uno_ampl.exe'.  Stop
```
This time, `libhighs` and the `cmake` subdirectory come from the `HiGHS_jll` artifact, correct?

.github/julia/build_tarballs_release.jl
```diff
@@ -105,7 +105,7 @@ cmake \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -DAMPLSOLVER=${libdir}/libasl.${dlext} \
-    -DHIGHS_DIR=${HIGHS_DIR} \
+    -DHIGHS=${HIGHS_DIR}/libhighs.${dlext} \
```
amontoison: @cvanaret `${HIGHS_DIR}` can be be replaced by `${libdir}` now.
We don't need to find `libhighs.dll.a` (import library) in ${prefix}/lib.
We only want the normal dll.

cvanaret:
We're good to go 😎

PR #512, opened 2026-02-14
[WIP] L-BFGS approximation for SQP methods

cvanaret:
Support for L-BFGS.

Limited memory quasi-Newton Hessian approximations will be computed based on the [compact QN representation](https://en.wikipedia.org/wiki/Compact_quasi-Newton_representation) (see Numerical optimization by Nocedal & Wright, pp 181-184).

@worc4021 @FrancoisGallard @leyffer 

Builds upon and supersedes #150.

- uses BLAS and LAPACK
- if no Hessian is provided for QPs and NPLs, pick L-BFGS Hessian

### DIFF
--- a/interfaces/Julia/test/NLP_wrapper.jl
+++ b/interfaces/Julia/test/NLP_wrapper.jl
@@ -32,6 +32,8 @@
     for j in 1:nlp.meta.ncon
         @test constraint_dual_solution[j] == UnoSolver.uno_get_constraint_dual_solution_component(solver, j-1)
     end
+
+    finalize(nlp)
 end
 
 @testset "uno_model -- uno_solver -- uno_optimize -- HS15 -- exact Hessian" begin


damienhocking:
Excellent

cvanaret:
@amontoison
Uno-filtersqp with a line search and an L-BFGS Hessian seems to run ✌️🎉
The code is neither very clean nor clever, I'll check it on Monday before I merge.

amontoison:
@cvanaret We already require `BLAS` in `Uno`, we should do the same thing with `LAPACK`.
Many libraries contain the symbols for both like MKL, OpenBLAS, ...

.github/julia/build_tarballs_release.jl
```diff
@@ -98,6 +98,7 @@ cmake \
     -DCMAKE_PREFIX_PATH=${libdir} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
+    -DWITH_LAPACK=ON \
```
amontoison: Please remove this option :)

.github/julia/runtests_uno_filtersqp_bqpd_ls_lbfgs.jl
```diff
@@ -0,0 +1,132 @@
...
+    Optimizer()
+
+Create a new `AmplNLWriter.Optimizer` object that uses Uno as the backing
+solver.
+"""
+function Optimizer(options)
+    return AmplNLWriter.Optimizer(Uno_jll.amplexe, options)
+end
+
+Optimizer_Uno_filtersqp() = Optimizer(["logger=SILENT", "preset=filtersqp", "QP_solver=BQPD", "globalization_mechanism=LS", "hessian_model=L-BFGS", "max_iterations=10000", "unbounded_objective_threshold=-1e15"])
```
amontoison: I suggest to use `LBFGS` instead of `L-BFGS` for the Hessian model.
I am not a big fan of `-` in it.

uno/ingredients/hessian_models/HessianModelFactory.cpp
```diff
@@ -8,12 +8,16 @@
 #include "ExactHessian.hpp"
 #include "IdentityHessian.hpp"
 #include "ZeroHessian.hpp"
+#ifdef HAS_LAPACK
```
amontoison: We should always require `LAPACK`.

uno/linear_algebra/LAPACK.hpp
```diff
@@ -0,0 +1,48 @@
...
+#ifndef UNO_LAPACK_H
+#define UNO_LAPACK_H
+
+#include "fortran_interface.h"
+#define LAPACK_cholesky_factorization FC_GLOBAL_(dpotrf, DPOTRF)
+#define LAPACK_symmetric_high_rank_update FC_GLOBAL_(dsyrk, DSYRK)
+#define LAPACK_triangular_back_solve FC_GLOBAL_(dtrsm, DTRSM)
+//#define LAPACK_matrix_matrix_product FC_GLOBAL_(dgemm, DGEMM)
+#define LAPACK_triangular_matrix_matrix_product FC_GLOBAL_(dtrmm, DTRMM)
+#define LAPACK_add_vector FC_GLOBAL_(daxpy, DAXPY)
```
amontoison: Why rename well-established routines that have been used for 50 years to names only Charlie understands? ;)
I suggest `uno_dpotrf`, `uno_dsyrk`, `uno_dtrsm`, `uno_dtrmm`, ...

amontoison: Why are you not using `dsytrf` and `dsytrs` as described in our document on Overleaf?

amontoison: Note that `axpy`, `trmm` and `trsm` are BLAS routines.
We should use a `BLAS.hpp` / `BLAS.cpp` fo them.

uno/linear_algebra/Vector.hpp
```diff
@@ -161,6 +161,15 @@ namespace uno {
...
    }
+
+   template <typename Vector, typename Vector2>
+   typename Vector::value_type dot(const Vector& x, const Vector2& y) {
+      typename Vector::value_type dot_product = 0;
+      for (size_t index: Range(x.size())) {
+         dot_product += x[index] * y[index];
+      }
+      return dot_product;
+   }
```
amontoison: Why not calling dot from BLAS?

uno/options/DefaultOptions.cpp
```diff
@@ -83,6 +83,9 @@ namespace uno {
       // use the primal-dual and dual step lengths to scale the dual directions when assembling the trial iterate
       options.set_bool("LS_scale_duals_with_step_length", true);
 
+      /* quasi-Newton options */
+      options.set_integer("quasi_newton_memory_size", 3);
```
amontoison: For information, the default value is `6` in Ipopt and `10` in KNITRO.

uno/options/Options.cpp
```diff
@@ -54,6 +54,7 @@ namespace uno {
       {"LS_backtracking_ratio", OptionType::DOUBLE},
       {"LS_min_step_length", OptionType::DOUBLE},
       {"LS_scale_duals_with_step_length", OptionType::BOOL},
+      {"quasi_newton_memory_size", OptionType::INTEGER},
```
amontoison, line 59: This new option should be documented in `options.md`.
We have a nice file with all options, we should do our best to keep it updated.

amontoison:
@cvanaret In the C interface, do you use LBFGS as fallback if the Hessian of the Lagrangian is not available?
Should we do that if the users didn't set `hessian_model=LBFGS`?

cvanaret:
> @cvanaret We already require `BLAS` in `Uno`, we should do the same thing with `LAPACK`. Many libraries contain the symbols for both like MKL, OpenBLAS, ...

At the moment, BLAS is only required if one of the linear solvers (HSL or MUMPS) is linked. Obviously it's always the case for `UnoSolver.jl` and `unopy`, but any C/C++ user who compiles with CMake might not link them. Should we systematically require BLAS and LAPACK anyway?

cvanaret:
> @cvanaret In the C interface, do you use LBFGS as fallback if the Hessian of the Lagrangian is not available? Should we do that if the users didn't set `hessian_model=LBFGS`?

I don't do it yet, but it's the plan yes!
At the moment we use a zero Hessian: https://github.com/cvanaret/Uno/blob/3c2eeae138e12d8f135fa319b2fdae4715de1b84/uno/ingredients/hessian_models/HessianModelFactory.cpp#L26

Edit: done

uno/options/Options.cpp
cvanaret: Oh yeah, thanks for reminding me we have that :sweat_smile:

uno/options/DefaultOptions.cpp
cvanaret: Thanks!

.github/julia/runtests_uno_filtersqp_bqpd_ls_lbfgs.jl
cvanaret: Works for me!

amontoison:
> > @cvanaret We already require `BLAS` in `Uno`, we should do the same thing with `LAPACK`. Many libraries contain the symbols for both like MKL, OpenBLAS, ...
> 
> At the moment, BLAS is only required if one of the linear solvers (HSL or MUMPS) is linked. Obviously it's always the case for `UnoSolver.jl` and `unopy`, but any C/C++ user who compiles with CMake might not link them. Should we systematically require BLAS and LAPACK anyway?

Yes, I think we should require them. We can rely more on it in the future this way. For example, the dot product can always be the one implemented in BLAS.

uno/ingredients/hessian_models/HessianModelFactory.cpp
cvanaret: Addressed in https://github.com/cvanaret/Uno/pull/513

.github/julia/build_tarballs_release.jl
cvanaret: Addressed in https://github.com/cvanaret/Uno/pull/513

uno/linear_algebra/Vector.hpp
cvanaret: Addressed in https://github.com/cvanaret/Uno/pull/513

uno/linear_algebra/LAPACK.hpp
cvanaret: > Note that `axpy`, `trmm` and `trsm` are BLAS routines. We should use a `BLAS.hpp` / `BLAS.cpp` fo them.

Addressed in https://github.com/cvanaret/Uno/pull/513

> Why are you not using `dsytrf` and `dsytrs` as described in our document on Overleaf?

So far, I simply took the 2025 PR which I completed based on the work of https://github.com/cvanaret/Uno/pull/497. I still need to optimize a few things.

cvanaret:
OK, I solved the failing workflows (C and Fortran) by linking LAPACK and BLAS.

uno/linear_algebra/LAPACK.hpp
cvanaret: > Why are you not using `dsytrf` and `dsytrs` as described in our document on Overleaf?

Ah, now that I look these up, I understand what you meant. I will merge this PR for SQP methods only, and I'll address the linear systems with quasi-Newton Hessians separately.

cvanaret:
Now I'm pretty proud of myself. I've implemented operator overloading/symbolic computations such that I can write the computation `U = (δ S + Y D⁻¹ Lᵀ) J⁻ᵀ` as:
```cpp
Uk = Sk;
Uk = this->delta * Uk + Vk * transpose(L_invsqrt_Dk);
Uk *= transpose(inverse(Jk));
```
and BLAS is called under the hood (`copy`, `gemm`, and `trsm`). Dimensions and leading dimensions are handled automatically.
Same for Cholesky (`potrf`) and low-rank update (`syrk`).
:star_struck:

amontoison:
@cvanaret Can you also add a small test for LBFGS in Fortran and Julia interfaces?
- https://github.com/cvanaret/Uno/blob/main/interfaces/Fortran/example_uno.f90
- https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/test/NLP_wrapper.jl

cvanaret:
> @cvanaret Can you also add a small test for LBFGS in Fortran and Julia interfaces?
> 
>     * https://github.com/cvanaret/Uno/blob/main/interfaces/Fortran/example_uno.f90
> 
>     * https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/test/NLP_wrapper.jl

@amontoison 
Done! I must have made a mistake (but which one?) when defining the test in `NLP_wrapper.jl`, I get:
```
Error During Test at /home/runner/work/Uno/Uno/interfaces/Julia/test/NLP_wrapper.jl:37
  Got exception outside of a @test
  CUTEst.CUTEstException(1, "memory allocation error")
```

amontoison:
You need the finalize the model if you recreate a CUTEstModel from the same problem.
You have an error because we don't dlopen / dlclose the shared library for the CUTEst problem the second time.

cvanaret:
@amontoison thanks, that's a wrap 😎

PR #513, opened 2026-02-15
BLAS and LAPACK

cvanaret:
- made BLAS and LAPACK mandatory in CMake 
- added interfaces `BLAS.hpp` and `LAPACK.hpp`
- wrote `dot()` with BLAS' `ddot`
- added functional tests
- updated action workflows

### DIFF
--- a/.github/workflows/build-msvc.yml
+++ b/.github/workflows/build-msvc.yml
@@ -25,6 +25,7 @@ jobs:
       matrix:
         os: [windows-latest]
         architecture: [x64]
+        library: [shared, static]
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


cvanaret:
@amontoison I'm close... I think the MSVC builds fail because the compilers (cl and ifx) are not the same as those of the artifacts?

amontoison:
If you don't force the compilation with gcc/gfortran/g++ on Windows, we need to recompile everything from scratch and we can't do that from the cross-compiler BinaryBuilder.

cvanaret:
> If you don't force the compilation with gcc/gfortran/g++ on Windows, we need to recompile everything from scratch and we can't do that from the cross-compiler BinaryBuilder.

@amontoison 
No problem. We have unit tests on MinGW anyway :) So if it's ok with you, I just delete the MSVC build workflow (edit: or I install BLAS and LAPACK via command line?)

amontoison:
I thought that it was for the Python bindings.
I just arrived at Chicago, I will have a look tonight.
If we just need to compile BLAS / LAPACK with MSVC, it should be fine.
My biggest fear is HiGHS on Windows and the various C++ runtime.

cvanaret:
@amontoison as soon as we link, the linker explodes. All the artifacts (except now BLAS and LAPACK) were compiled with `gfortran`, right? Should we just compile the static `libuno`?

uno/linear_algebra/BLAS.hpp
```diff
@@ -0,0 +1,39 @@
...
+   // x^T y
+   double BLAS_dot_product(const int* n, const double* x, const int* incx, const double* y, const int* incy);
+
+   // performs back-solve with lower-triangular A:
+   // op(A) X = alpha B    or
+   // X op(A) = alpha B
+   // where
+   // op(A) = A   or   op(A) = A^T
+   // X is overwritten on B
+   void BLAS_triangular_back_solve(char* side, char* uplo, char* transa, char* diag, int* m, int* n, double* alpha,
```
amontoison, line 27: I suggest to add `matrix` in the name of the routine because it is `trsm` and not `trsv`.

amontoison: Can be done in the future...

PR #514, opened 2026-02-16
[CI] Add a common script unit-tests.yml

amontoison:
- One script for linux / mac / windows
- Add tests for mac x64 and linux aarch64

PR #515, opened 2026-02-16
[CI] Add a common script build.yml

amontoison:
Similar to #514 for `build-*.yml`


amontoison:
I needed to recompile HiGHS without ZLIB...
We should document how to compile Uno with `clang / clang++` if a user wants a shared library `libuno.dylib`.
It is non-trivial :skull:

PR #518, opened 2026-02-16
Upgrade to HiGHS 1.14.0

amontoison:
Charlie, we have an issue with HiGHS 1.13.0 and 1.13.1.
We have a `free(): invalid pointer`...


cvanaret:
Not a very productive answer, but it works fine on my Linux machine :sweat_smile: (HiGHS 1.13.1 compiled from source)

cvanaret:
Things look good with HiGHS v1.14.0. The issue is that METIS is now part of HiGHS, so the symbols of libmetis and libhighs clash.

PR #534, opened 2026-02-27
[unopy] Release v0.2.1 (failed to merge)

cvanaret:


cvanaret:
Note: did not trigger upload to PyPI when merged. Reason: since there is a single commit, the title merge commit was that of the commit in the PR.
I fixed this in the settings, should not happen again. Re-triggered in https://github.com/cvanaret/Uno/pull/535

PR #538, opened 2026-02-27
[UnoSolver.jl] Pass the preset before the other options

### LINKED ISSUE #537: [UnoSolver.jl] preset may overwrite other options in `MOI_Wrapper.jl`
The option dictionary in `MOI_Wrapper.jl` does not preserve the insertion order. Consequently, the dictionary built from the kwargs `UnoSolver.Optimizer(preset="filtersqp", hessian_model="LBFGS", globalization_mechanism="LS")` ends up looking like:
```julia
Dict{String, Any} with 3 entries:
  "globalization_mechanism" => "LS"
  "hessian_model"           => "LBFGS"
  "preset"                  => "filtersqp"
```

and the preset overwrites the two previous options.
@amontoison Should be simply handle the preset separately at the beginning?

amontoison:
close #537

### DIFF
--- a/interfaces/Julia/src/C_wrapper.jl
+++ b/interfaces/Julia/src/C_wrapper.jl
@@ -246,7 +246,7 @@ function uno_solver(; kwargs...)
 
   # pass options to Uno
   if haskey(kwargs, :preset)
-    uno_set_solver_preset(solver, kwargs.preset)
+    uno_set_solver_preset(solver, kwargs[:preset])
   end
   for (k, v) in kwargs
     (k == :preset) && continue


interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1388,8 +1388,14 @@ function MOI.optimize!(model::Optimizer)
     # The default logger is "INFO".
     UnoSolver.uno_set_solver_string_option(solver, "logger", model.silent ? "SILENT" : "INFO")
 
+    # If provided, set the preset before any other options.
+    if haskey(model.options, "preset")
+        UnoSolver.uno_set_solver_preset(solver, model.options[preset])
```
amontoison:
```suggestion
        UnoSolver.uno_set_solver_preset(solver, model.options["preset"])
```

cvanaret:
Looks great @amontoison!
Before I merge, here's a quick question: since the MOI option dictionary is of type `Dict{String,Any}`, I could store with the value the current index of the kwarg. Something like:
```julia
option_dict = Dict{String,Any}()
current_index = 0
for (name, value) in kwargs
    option_dict[name |> string] = (value, current_index)
    current_index = current_index + 1
end
```
then I would sort `option_dict` wrt the indices before setting the solver options, so that we traverse the elements in the right order:

```julia
sort(collect(option_dict), by = x -> x[2][2])
```

In most cases the two solutions are equivalent, but in the cases where the user does something stupid, e.g.,
```julia
UnoSolver.Optimizer(globalization_mechanism="LS", preset="filtersqp", hessian_model="LBFGS")
```
(globalization_mechanism="LS" is overwritten by the preset's "TR"), we still want a predictable behavior and respect the user's order.

What do you think? I'm happy to give it a try.

amontoison:
Except the `preset`, which we can explictly add as a field in the `Solver` or `Optimizer`, do we have any other options where the order is critical?

cvanaret:
Fair point. I'll merge and if someone complains someday, we can reconsider.

PR #543, opened 2026-03-03
[unopy] Release v0.2.2 (fixed macOS wheels)

cvanaret:
Build macOS wheels with clang/clang++/gfortran instead of full GNU

cc @amontoison 

Ok now we're using the right compilers. Shouldn't be too hard.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ build-backend = "scikit_build_core.build"
 
 [project]
 name = "unopy"
-version = "0.2.1"
+version = "0.2.2"
 #requires-python = ">=3.8"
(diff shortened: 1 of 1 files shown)


amontoison:
@cvanaret 
https://github.com/cvanaret/Uno/blob/main/.github/workflows/build.yml

cvanaret:
> @cvanaret https://github.com/cvanaret/Uno/blob/main/.github/workflows/build.yml

Exactly! I'll do that when I'm at work.

cvanaret:
It ended up being more involved than I thought, but it finally works. Thanks Claude!

Issue #544, opened 2026-03-03
[UnoSolver.jl] LBFGS return `Algorithmic error`


franckgaga:
First many thanks for your work on the new builtin LBFGS method!

I experiment with it on a simple problem (which is convex in theory, so it should be easily solved by the quasi-newton approach) using ModelPredictiveControl.jl, and JuMP returns termination status `OTHER_ERROR`. The raw status of `UnoSolver.jl` is `4`, and the logging shows `Algorithmic error`.

My problem works well if replace `hessian=false` with `hessian=true`, to provide an exact Hessian to `UnoSolver.jl`.

Here's my MRE:

```julia
using ModelPredictiveControl
using JuMP, UnoSolver
N = 1
f!(ẋ, x, u, _ , _ ) = (ẋ .= -0.1.*x .+ u)
h!(y, x, _ , _ ) = (y .= x)
model = NonLinModel(f!, h!, 1.0, 1, 1, 1)
optim = Model(()->UnoSolver.Optimizer(
    preset="filtersqp", hessian_model="LBFGS", globalization_mechanism="LS"
))
hessian = false
nmpc = NonLinMPC(model; Hp=20, oracle=true, hessian, optim)
unset_time_limit_sec(nmpc.optim)
unset_silent(nmpc.optim)
res = sim!(nmpc, N)
display(getinfo(nmpc)[:sol])
using Plots; display(plot(res))
```
printing:
```
Original model C model
3 variables, 1 constraints (0 equality, 1 inequality)

Non-default options:
LP_solver = BQPD

───────────────────────────────────────────────────────────────────────────────────────────────────────────
  Iterations
 Major  Minor  Penalty   Steplength Phase ||Step||  Objective  Infeas    Statio    Compl     Status        
───────────────────────────────────────────────────────────────────────────────────────────────────────────
 0      -      -         -          OPT   -         2.000e+01  0.00e+00  3.21e+02  0.00e+00  initial point 
 -      -      -         -          -     -         -          -         -         -         The subproblem is unbounded, this should not happen. If the subproblem has curvature,use regularization. If not, use a trust-region method.

───────────────────────────────────────────────────────────────────────────────────────────────────────────
  Iterations
 Major  Minor  Penalty   Steplength Phase ||Step||  Objective  Infeas    Statio    Compl     Status        
───────────────────────────────────────────────────────────────────────────────────────────────────────────

[...]

PR #545, opened 2026-03-03
[UnoSolver.jl] Support LBFGS with MathOptInterface / JuMP

amontoison:
Removed redundant check for `hessian_sparsity` when setting problem_type to `'LP'`.

Related issue: #544

### DIFF
--- a/interfaces/Julia/src/C_wrapper.jl
+++ b/interfaces/Julia/src/C_wrapper.jl
@@ -230,6 +230,11 @@ mutable struct Solver
   c_solver::Ptr{Cvoid}
 end
 
+function uno_get_solver_string_option(solver::Solver, option_name::String)
+  option_value = uno_get_solver_string_option(solver.c_solver, option_name)
+  return unsafe_string(option_value)
+end
+
 function uno_destroy_solver(solver::Solver)
   if solver.c_solver != C_NULL
     uno_destroy_solver(solver.c_solver)
--- a/interfaces/Julia/test/C_wrapper.jl
+++ b/interfaces/Julia/test/C_wrapper.jl
@@ -227,6 +227,7 @@ y0 = zeros(Float64, ncon)
   solution_primal_feasibility = UnoSolver.uno_get_solution_primal_feasibility(solver)
   solution_stationarity = UnoSolver.uno_get_solution_stationarity(solver)
   solution_complementarity = UnoSolver.uno_get_solution_complementarity(solver)
+  hessian_model = UnoSolver.uno_get_solver_string_option(solver, "hessian_model")
 
   primal_solution = Vector{Float64}(undef, nvar)
   UnoSolver.uno_get_primal_solution(solver, primal_solution)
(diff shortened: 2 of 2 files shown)


amontoison:
@cvanaret Can you check why some MOI tests are failing?
I will be offline the next 4 hours 😁

cvanaret:
It messed up some dual signs that were relying on the correction:
```julia
function _dual_multiplier(model::Optimizer)
    return xor(model.problem_type == "LP", model.sense == MOI.MAX_SENSE) ? 1.0 : -1.0
```
:scream:

interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1673,7 +1670,8 @@ end
 ### MOI.ConstraintDual
 
 function _dual_multiplier(model::Optimizer)
-    return xor(model.problem_type == "LP", model.sense == MOI.MAX_SENSE) ? 1.0 : -1.0
+    return 1.
+    #return xor(model.problem_type == "LP", model.sense == MOI.MAX_SENSE) ? 1.0 : -1.0
```
amontoison:
```suggestion
    return model.sense == MOI.MAX_SENSE ? 1.0 : -1.0
```

interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1672,8 +1669,16 @@ end
 
 ### MOI.ConstraintDual
 
+function _quasi_newton_approximation(model::Optimizer)::Bool
+    hessian_model = UnoSolver.uno_get_solver_string_option(model.solver, "hessian_model")
+    nlp_without_hessian = (model.problem_type != "LP") && (length(model.hrows) == 0)
```
amontoison, line 1674: @cvanaret Do you set the option `hessian_model = LBFGS` for the user if the operator `Hv` / sparse Hessian `H` is not provided?
I know that you use `LBFGS` if nothing else is available but I am not sure that you also update the value of the option `"hessian_model"`.

amontoison: I checked and the answer is no, you don't change manually the option `hessian_model` in the C interface.
You use L-BFGS directly in the C++ code.

I need to keep the second check.

cvanaret: Ah! I never thought the user might want to know which options are overridden by Uno :scream: Makes total sense. I'll implement that tomorrow (https://github.com/cvanaret/Uno/issues/559).

amontoison: I will do a new release of `UnoSolver.jl` with the current fix.
A new release `Uno_jll` will be needed for what you plan to implement for #559.

Issue #484, opened 2026-01-10
Unovector __setitem__ in pybind11 only supports element wise set


FrancoisGallard:
To fill the Jacobians we have to set each index of the Unovector binded in Python. This is very slow and wont scale with many design variables or constraints.
We should be able to set the values of a UnoVector from a memory contiguous numpy array or a ctype.
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html
https://docs.python.org/3/library/ctypes.html#arrays

PR #548, opened 2026-03-03
[unopy] Improve pybind11 bindings

cvanaret:
- [x] replaced `PointerWrapper` with native `py::array_t`
- [x] converted `uno::Vector` in `Result` to `py::array_t`
- [x] moved the arrays and callbacks into `Model` instead of copying
- [x] removed useless parameters (sizes, user_data) in callbacks
- [x] have the callbacks throw exceptions, instead of returning an error code 

cc @FrancoisGallard

Fixes https://github.com/cvanaret/Uno/issues/484

### DIFF
--- a/interfaces/Python/example/example_hs015.py
+++ b/interfaces/Python/example/example_hs015.py
@@ -12,34 +12,30 @@ def objective(x):
 	
 def constraints(x, constraint_values):
 	# the modeler should throw an exception if the function cannot be evaluated
-	constraint_values[0] = x[0]*x[1]
-	constraint_values[1] = x[0] + x[1]**2
+	constraint_values[:] = [x[0]*x[1], x[0] + x[1]**2]
 	
 def objective_gradient(x, gradient):
 	# the modeler should throw an exception if the function cannot be evaluated
-	gradient[0] = 400.*x[0]**3 - 400.*x[0]*x[1] + 2.*x[0] - 2.
-	gradient[1] = 200.*(x[1] - x[0]**2)
+	gradient[:] = [400.*x[0]**3 - 400.*x[0]*x[1] + 2.*x[0] - 2.,
+				   200.*(x[1] - x[0]**2)]
 
 def jacobian(x, jacobian_values):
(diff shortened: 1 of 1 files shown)


cvanaret:
Hi @FrancoisGallard,
I've made good progress here, unopy is now compatible with numpy arrays (https://github.com/cvanaret/Uno/issues/484). The last thing I would do before merging would be to replace the existing syntax:

```python
def objective(x, objective_value):
	objective_value[0] = 100.*(x[1] - x[0]**2)**2 + (1. - x[0])**2
	return 0
	
def constraints(x, constraint_values):
	constraint_values[0] = x[0]*x[1]
	constraint_values[1] = x[0] + x[1]**2
	return 0
```
with something like:
```python
def objective(x):
	# if evaluation error, throw some exception
	return 100.*(x[1] - x[0]**2)**2 + (1. - x[0])**2
	
def constraints(x, constraint_values):
	# if evaluation error, throw some exception
	constraint_values[0] = x[0]*x[1]
	constraint_values[1] = x[0] + x[1]**2
```
Do you have objections/suggestions?

robfalck:
Does Uno have any provision for detecting invalid design points / backtracking?

If so then maybe the API should be more like the original, using the return to specify success/failure.

```python
def objective(x, objective_value):
    """
	objective_value[0] = 100.*(x[1] - x[0]**2)**2 + (1. - x[0])**2
	return True
```

And just to demonstrate that non-elemental access works:

```python
def constraints(x, constraint_values):
	constraint_values[:] = [ x[0]*x[1], x[0] + x[1]**2 ]
	return True
```

cvanaret:
@robfalck yes, Uno backtracks when the trial iterate cannot be evaluated (at the moment, this is true only for objective and constraints, but I'm planning to look ahead at the derivatives too).

I'm fine with the callbacks returning a boolean (like we have now) or throwing an exception (which I find more elegant).

robfalck:
Agreed that requiring an exception to be raised would be fine.

PR #480, opened 2025-12-16
[unopy] build wheels for Windows

cvanaret:

PR #549, opened 2026-03-04
[unopy] Test the TestPyPI artifacts

cvanaret:


amontoison:
The issue for Ubuntu is related to `cxx11` string ABI:
https://docs.binarybuilder.org/stable/tricksy_gotchas/

We will also get the same error on windows (if we found a way to use unopy with CI on this platform).
I need to recompile `UnoUtils`.

cvanaret:
@amontoison how come the current artifacts from PyPI work fine, but the new TestPyPI artifacts fail? Did we change anything that would cause this cxx11 string ABI issue?

amontoison:
I think it was always there.
But I checked the symbols in the precompiled library of `UnoUtils` and nothing is wrong.
The issue is with `cibuildwheel`.
The flag `-D_GLIBCXX_USE_CXX11_ABI=0` is forced somewhere while we want `-D_GLIBCXX_USE_CXX11_ABI=1`.

cvanaret:
I see. Note: the TestPyPI artifacts were built with the cibuildwheel action, I'm now rebuilding in https://github.com/cvanaret/Uno/pull/480 with a direct call `cibuildwheel --output-dir ${{github.workspace}}/wheelhouse`, like we have in the main branch. Could be the source of the problem.

amontoison:
The good news is that now I understand why we can't do a static linking of HiGHS on Windows, we have the same issue on Linux but the linker is more permissive and we only have the error at runtime.

amontoison:
I also suggest to test all platforms (aarch64 linux and x64 mac are missing).

amontoison:
@cvanaret If it can help:
https://github.com/ralna/GALAHAD/commit/95441a46a542c2defb594fbd9112f42f3029f661#diff-6b39a2f1a6399aa39e427fb4c11a7b552e615a2d83a3c36aa6dd6ddbbb8281f4

I remember that Jari spend a lot of time on this issue.

cvanaret:
@amontoison cool, that's exactly what Claude asked me to add to `__init__py` (https://github.com/cvanaret/Uno/pull/550/changes/33502d413f11028e3efaf5d58a51b810eadd998e) :smiley:

cvanaret:
@amontoison good news! If we exclude `libhighs.dll` from the wheels, unopy runs on Windows :)
https://github.com/cvanaret/Uno/actions/runs/22777727310/job/66075263282?pr=549

amontoison:
We can maybe do a release of unopy for Windows without it and add it later?
We still need to understand why HiGHS 1.13 is failing with Uno

cvanaret:
> We can maybe do a release of unopy for Windows without it and add it later? We still need to understand why HiGHS 1.13 is failing with Uno

Agreed. Then we can finally merge https://github.com/cvanaret/Uno/pull/480. I'll tag you when it's ready.

amontoison:
@cvanaret Should we close this PR?

cvanaret:
@amontoison there is useful code to detect DLL errors in wheels. I would keep it open in case we need it when we add HiGHS.

PR #561, opened 2026-03-08
[UnoSolver.jl] Improve the support for unconstrained problems

### LINKED ISSUE #464: [UnoSolver.jl] Add a more-user friendly function uno_model for unconstrained problems
https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/src/C_wrapper.jl#L112-L140

amontoison:
close #464
- Add new methods `uno_model` and `uno` for unconstrained problems in the low-level interface of `UnoSolver.jl`. 
- Update the extension for `NLPModels.jl`.

### DIFF
--- a/interfaces/Julia/src/C_wrapper.jl
+++ b/interfaces/Julia/src/C_wrapper.jl
@@ -225,7 +225,6 @@ function uno_model(
   eval_Hv::Union{Function,Nothing},
   user_model=nothing,
   hessian_triangle::Char='L',
-  lagrangian_sign::Float64=1.0,
   x0::Union{Vector{Float64},Nothing}=nothing,
 )
   return uno_model(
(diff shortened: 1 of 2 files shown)


interfaces/Julia/src/C_wrapper.jl
```diff
@@ -193,20 +196,69 @@ function uno_model(
...
+  hrows::Vector{Cint},
+  hcols::Vector{Cint},
+  nnzh::Int,
+  eval_objective::Function,
+  eval_gradient::Function,
+  eval_hessian::Union{Function,Nothing},
+  eval_Hv::Union{Function,Nothing},
+  user_model=nothing,
+  hessian_triangle::Char='L',
+  lagrangian_sign::Float64=1.0,
```
cvanaret: `lagrangian_sign` (at least for the Lagrangian Hessian) is irrelevant for unconstrained problems.

amontoison: Good catch, I fixed it and added a test with an unconstrained problem (`hs5`).

PR #541, opened 2026-03-02
[WIP] Interface to SPRAL's SSIDS linear solver

cvanaret:
- [x] SSIDS interface 
- [x] document the fix for the lag at startup

Fixes https://github.com/cvanaret/Uno/issues/284

PR #569, opened 2026-03-10
[UnoSolver.jl] Release Set OpenMP environment variables in workflow

amontoison:
Update: It will be needed for the next release with #541

### DIFF
--- a/.github/julia/build_tarballs_utils.jl
+++ b/.github/julia/build_tarballs_utils.jl
@@ -1,8 +1,11 @@
 # julia +1.7 --color=yes build_tarballs_utils.jl x86_64-linux-gnu-libgfortran5-cxx11,x86_64-apple-darwin-libgfortran5-cxx11,x86_64-w64-mingw32-libgfortran5-cxx11,aarch64-linux-gnu-libgfortran5-cxx11,aarch64-apple-darwin-libgfortran5-cxx11 --verbose --deploy="amontoison/UnoUtils_jll.jl"
+#
+# Note: the following variables is needed for cross-compilation on Mac platforms in the .bashrc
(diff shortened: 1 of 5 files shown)


amontoison:
@cvanaret It will be great to not seg fault if we do a mistake in the name of the linear solver.
`SSIDS` was not added in the release 2.4.1 but if we try to use it, Uno returns:
```julia
Solving with linear solver SSIDS
libc++abi: terminating due to uncaught exception of type std::invalid_argument
...
```

amontoison:
@cvanaret The tests passed on Windows :thinking: 
It seems that the linear solver is not modified on Windows :cry:

cvanaret:
> @cvanaret It will be great to not seg fault if we do a mistake in the name of the linear solver. `SSIDS` was not added in the release 2.4.1 but if we try to use it, Uno returns:
> 
> ```julia
> Solving with linear solver SSIDS
> libc++abi: terminating due to uncaught exception of type std::invalid_argument
> ...
> ```

That's weird. SSIDS is not in Uno 2.4.1, simply merged in main. And if we make a mistake in the name of the linear solver, it should terminate graciously and print the names of the available solvers.

cvanaret:
> @cvanaret It will be great to not seg fault if we do a mistake in the name of the linear solver. `SSIDS` was not added in the release 2.4.1 but if we try to use it, Uno returns:
> 
> ```julia
> Solving with linear solver SSIDS
> libc++abi: terminating due to uncaught exception of type std::invalid_argument
> ...
> ```

Ah wait, I know where it comes from. For some reason I have a safeguard in the AMPL interface (a relic of the first Uno) that hides the thrown exception. I'm on it (https://github.com/cvanaret/Uno/pull/572).

amontoison:
Done in #586

Issue #566, opened 2026-03-09
Implement limited-memory SR1


cvanaret:
SR1 is a quasi-Newton Hessian approximation that captures negative curvature, contrary to BFGS. The [limited-memory SR1](https://en.wikipedia.org/wiki/Symmetric_rank-one#Limited_Memory) will be implemented with the [compact representation](https://en.wikipedia.org/wiki/Compact_quasi-Newton_representation).

@amontoison the paper *Representations of quasi-Newton matrices and their use in limited memory methods* (link [here](https://web.archive.org/web/20170808213836id_/http://www.ii.uib.no/~lennart/drgrad/Byrd1994.pdf)) describes the compact representation, and states on p149:

> when using the form (5.2) in a limited memory method, the condition (5.10) could be tested when computing a **triangular factorization of $M_k$ without pivoting**, with the test for a zero on the diagonal of the factor being made relative to the magnitude of $Q_k$ and $S_k$. Skipping an update would correspond to deleting the corresponding row and column of $M_k$.

where $M_k = D_{k}+L_{k}+L_{k}^{T}-S_{k}^{T}B_{0}S_{k}$ ( = Wikipedia's $N_k$) is indefinite. Which LAPACK routine would you call to compute this factorization without pivoting?

amontoison:
@cvanaret To the best of my knowledge, reference LAPACK doesn't have an LDL' without pivoting.
Without pivoting you only have the Cholesky (`potrf`) or the LU (`getrfnp`), where `np = no-pivoting`.
For symmetric indefinite matrices, we only have `sytrf` but it is LBL' with pivoting.

I think we need to implement our own dense LDL'.

@langou I remember that we talked about it during my visit to Denver, can you confirm?

langou:
> To the best of my knowledge, reference LAPACK doesn't have an LDL\' without pivoting. Without pivoting you only have the Cholesky ('potrf') or the LU ('getrfnp'), where 'np = no-pivoting'. For symmetric indefinite matrices, we only have 'sytrf' but it is LBL' with pivoting.

@amontoison is correct. 

This is not too hard to write. You should be able to get reasonable performance by writing a recursive implementation. 

If there is interest in an LDL\' without pivoting, we could add it to the wishlist of LAPACK.

cvanaret:
@amontoison @langou here's a (C++) draft of the factorization and solve routines.

```c++
// Partition A (n×n, column-major with leading dimension lda, symmetric) as
//
[...]

PR #577, opened 2026-03-11
[WIP] L-SR1 Hessian approximation

cvanaret:
[L-SR1](https://en.wikipedia.org/wiki/Symmetric_rank-one#Limited_Memory) is a limited-memory, indefinite quasi-Newton Hessian approximation. It is best paired with a trust-region method to exploit negative curvature.
Represented implicitly via the [compact representation](https://en.wikipedia.org/wiki/Compact_quasi-Newton_representation). Runs on BLAS and LAPACK.

Closes https://github.com/cvanaret/Uno/issues/566

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Uno unifies Lagrange-Newton (essentially **SQP** and **interior-point**) methods
 Uno currently implements the following strategies:
 * **constraint relaxation strategies**: feasibility restoration;
 * **inequality handling methods**: inequality constrained method, interior-point method;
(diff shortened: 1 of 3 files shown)


cvanaret:
OK the CI tests pass :muscle:

amontoison:
@cvanaret is it only for `filtersqp` or is it also working in the preset `ipopt`?

cvanaret:
It should be used with a trust-region method because the L-SR1 Hessian approximation may be indefinite.

PR #578, opened 2026-03-11
[C/UnoSolver.jl] Properly set model properties for LPs

cvanaret:
@amontoison I'm investigating why adding L-BFGS suddenly broke the dual signs in the `UnoSolver.jl` tests (before you fixed it).
There are instances like [this one](https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Test/test_nonlinear.jl#L1626). This is apparently an LP but registered with `ScalarNonlinearFunction`, so it's flagged as NLP. No Hessian information is available though (that's why now Uno is using L-BFGS instead of no Hessian). How come? Obviously this is an LP, but we should at least get a 0 Hessian from MOI.

### DIFF
--- a/interfaces/Julia/test/MOI_wrapper.jl
+++ b/interfaces/Julia/test/MOI_wrapper.jl
@@ -28,7 +28,7 @@ function test_MOI_Test()
         with_cache_type = Float64,
         with_bridge_type = Float64,
     )
-    # MOI.set(model, MOI.Silent(), true)
+    MOI.set(model, MOI.Silent(), true)
     MOI.Test.runtests(
         model,
         MOI.Test.Config(
@@ -69,7 +69,7 @@ end
 
 function test_ConstraintDualStart()
     model = UnoSolver.Optimizer()
-    # MOI.set(model, MOI.Silent(), true)
+    MOI.set(model, MOI.Silent(), true)
     x = MOI.add_variables(model, 2)
     c = MOI.add_constraint(
         model,
@@ -93,7 +93,7 @@ end
 
 function test_ConstraintDualStart_ScalarNonlinearFunction()
     model = UnoSolver.Optimizer()
-    # MOI.set(model, MOI.Silent(), true)
+    MOI.set(model, MOI.Silent(), true)
     x = MOI.add_variables(model, 2)
     MOI.add_constraint.(model, x, MOI.Interval(0.0, 0.8))
     f = MOI.ScalarNonlinearFunction(:sin, Any[1.0*x[1]-1.0*x[2]])
@@ -115,7 +115,7 @@ end
 
 function test_solve_time()
     model = UnoSolver.Optimizer()
-    # MOI.set(model, MOI.Silent(), true)
+    MOI.set(model, MOI.Silent(), true)
     MOI.add_variable(model)
     @test isnan(MOI.get(model, MOI.SolveTimeSec()))
     MOI.optimize!(model)
@@ -125,7 +125,7 @@ end
 
 function test_barrier_iterations()
     model = UnoSolver.Optimizer()
-    # MOI.set(model, MOI.Silent(), true)
+    MOI.set(model, MOI.Silent(), true)
     x = MOI.add_variable(model)
     f = (x - 1.0)^2 + 2.0 * x + 3.0
     MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
@@ -152,7 +152,7 @@ end
 
 function test_empty_optimize()
     model = UnoSolver.Optimizer()
-    # MOI.set(model, MOI.Silent(), true)
+    MOI.set(model, MOI.Silent(), true)
     @test MOI.get(model, MOI.RawStatusString()) == "Optimize not called"
     MOI.optimize!(model)
     @test MOI.get(model, MOI.TerminationStatus()) == MOI.INVALID_MODEL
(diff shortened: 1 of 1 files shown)


amontoison:
It is just a unit test and Oscar probably forced it as an NLP with the low-level MOI API for coverage reasons.
For example it could test that the solver is not failing if `nnzh = 0`.
If you use the JuMP API it is should be always correctly flagged as a LP.

cc @odow for confirmation

odow:
JuMP/MOI has a very simple dual convention:

 * if the set is GreaterThan, then the dual is non-negative
 * if the set is LessThan, then the dual is non-positive

It doesn't depend on the objective sense. It doesn't depend on the problem type. It doesn't depend whether the function is ScalarAffineFunction of ScalarNonlinearFunction.

The tests explicitly include a bunch of edge cases to catch out solvers. If you fail the test, it's either a bug in the wrapper or a bug in the solver.

> If you use the JuMP API it is should be always correctly flagged as a LP.

Not necessarily:

```Julia
julia> begin
           model = Model()
           @variable(model, x)
           @constraint(model, @force_nonlinear(x + 1) <= 1.0)
           model
       end
A JuMP Model
├ solver: none
├ objective_sense: FEASIBILITY_SENSE
├ num_variables: 1
├ num_constraints: 1
│ └ NonlinearExpr in MOI.LessThan{Float64}: 1
└ Names registered in the model
  └ :x

julia> print(model)
Feasibility
Subject to
 (x + 1) - 1 ≤ 0
```

cvanaret:
Thanks @odow!
@amontoison what I don't get is why our MOI interface gives us an NLP without Hessian information here.

amontoison:
@cvanaret 
Can you print `has_hessian` after this line to confirm if the issue is in the Julia interface?https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl#L1281

If you provide a Hessian of Lagrangian with `nnzh = 0`, does Uno automatically change the problem type for `"LP"` ?

cvanaret:
@amontoison `is_hessian` is true, but the Hessian is never passed to the model because `nnz = 0` (https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/src/C_wrapper.jl#L167).
No, Uno doesn't modify the problem type.

So your fix of the dual signs [here](https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl#L1679) actually captures this particular scenario (NLP without Hessian information).

amontoison:
> @amontoison `is_hessian` is true, but the Hessian is never passed to the model because `nnz = 0` (https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/src/C_wrapper.jl#L167). No, Uno doesn't modify the problem type.

Good catch!
We should change that to always pass the Hessian of the Lagrangian / Hessian operator even if `nnzh == 0` (remove this check).
It is also a very bad check because I use the same condition to pass the Hessian operator and if we only want to pass it, it is relevant to keep empty vectors for `hrows` / `hcols` and set `nnzh = 0` in the Julia parameters.

> So your fix of the dual signs [here](https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl#L1679) actually captures this particular scenario (NLP without Hessian information).

Good but we can improve the detection of the problem type for the user.
If `nnzh == 0`, the user provides the Hessian of the Lagrangian, and `problem_type = "NLP"`, we can probably override `problem_type` in the C structure of the Options.

cvanaret:
> > @amontoison `is_hessian` is true, but the Hessian is never passed to the model because `nnz = 0` (https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/src/C_wrapper.jl#L167). No, Uno doesn't modify the problem type.
> 
> Good catch! We should change that to always pass the Hessian of the Lagrangian / Hessian operator even if `nnzh == 0` (remove this check). It is also a very bad check because I use the same condition to pass the Hessian operator and if we only want to pass it, it is relevant to keep empty vectors for `hrows` / `hcols` and set `nnzh = 0` in the Julia parameters.
> 
> > So your fix of the dual signs [here](https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_wrapper.jl#L1679) actually captures this particular scenario (NLP without Hessian information).
> 
> Good but we can improve the detection of the problem type for the user. If `nnzh == 0`, the user provides the Hessian of the Lagrangian, and `problem_type = "NLP"`, we can probably override `problem_type` in the C structure of the Options.

Works for me :)

interfaces/C/Uno_C_API.cpp
```diff
@@ -574,7 +574,7 @@ bool uno_set_jacobian_transposed_operator(void* model, JacobianTransposedOperato
 bool uno_set_lagrangian_hessian(void* model, uno_int number_hessian_nonzeros, char hessian_triangular_part,
       const uno_int* hessian_row_indices, const uno_int* hessian_column_indices, Hessian lagrangian_hessian,
       double lagrangian_sign_convention) {
-   if (number_hessian_nonzeros <= 0) {
+   if (number_hessian_nonzeros < 0) {
```
amontoison, line 577: It is because of this condition that I added `nnzh > 0` in the Julia interface!
It was failing when a Hessian was provided for LP problems.

cvanaret: Yeah I had forgotten about it! It should be nice and consistent now.

amontoison:
@cvanaret Can you also backport my modifications in the Julia interface to make everything more robust?
--> https://github.com/cvanaret/Uno/compare/main...am/nnzh_check

PR #597, opened 2026-03-17
Do not perform the first iteration if `max_iterations == 0`

cvanaret:
See https://github.com/control-toolbox/CTSolvers.jl/pull/103#discussion_r2945056681


cvanaret:
@amontoison in `interfaces/Julia/test/MOI_wrapper.jl`, the test `test_vector_nonlinear_oracle` contains:
```julia
MOI.set(model, MOI.RawOptimizerAttribute("max_iterations"), 0)
```
Is that normal?

amontoison:
It comes from this test in `Ipopt.jl`:
https://github.com/jump-dev/Ipopt.jl/pull/511

You can modify / remove it if it is not relevant for Uno.

cvanaret:
Got it, thanks. Uno projects the initial point onto the bounds, so this test isn't relevant :+1:

PR #645, opened 2026-04-04
Functional tests for linear solvers

cvanaret:
- Re-activated functional tests for MA57, MA27, and MUMPS
- Added functional tests for SSIDS


cvanaret:
Hi @amontoison,
I still cannot build the unit tests for MinGW because `libmetis.dll` is not on the PATH for the target `run_unotest`:

```
cygcheck: track_down: could not find libmetis.dll
```

At the moment we're linking `dependencies\bin\libmetis.dll`. Could I link `dependencies\lib\libmetis.a` or `dependencies\lib\libmetis.dll.a` instead?

cvanaret:
@amontoison I ended up installing the MSYS2 METIS (`mingw-w64-x86_64-metis`) and it works :slightly_smiling_face:

PR #657, opened 2026-04-14
[CI] Added build for Ubuntu + Intel compilers

cvanaret:


cvanaret:
@amontoison I had added this workflow to check an issue that is now solved. Is it interesting to keep or can I drop it?

PR #675, opened 2026-04-23
Pass the bounds separately during model creation

### LINKED ISSUE #671: [APIs] Pass the variables bounds separately at model creation
We should consider passing the variables bounds separately at model creation. At the moment, we have in the C API:
```c
void* uno_create_model(const char* problem_type, uno_int number_variables, const double* variables_lower_bounds,
      const double* variables_upper_bounds, uno_int base_indexing);
```

I'm suggesting:
```c
void* uno_create_model(const char* problem_type, uno_int number_variables, uno_int base_indexing);

void uno_set_variables_bounds(void* model, const double* variables_lower_bounds, const double* variables_upper_bounds);
```

This would be beneficial for:
- the CasADi interface (cc @david0oo) because the variables bounds are known only when the problem is solved
- reoptimization with updated variables bounds (e.g., within an MINLP framework, see https://github.com/cvanaret/Uno/issues/107)

This would obviously be a breaking change in all APIs. @amontoison any objections?

cvanaret:
- Added `uno_create_unconstrained_model()`, `uno_set_variables_lower_bounds()`, `uno_set_variables_upper_bounds()`, `uno_set_variable_lower_bound()` and `uno_set_variable_upper_bound()` to:
  - [x] C API
  - [x] Julia interface
  - [x] Python interface
  - [x] Fortran interface

- Have `uno_create_model()` call both `uno_create_unconstrained_model()`, `uno_set_variables_lower_bounds()` and `uno_set_variables_upper_bounds()` in C API

- Other routines added: `uno_set_constraints_lower_bounds()`, `uno_set_constraints_upper_bounds()`, `uno_set_constraint_lower_bound()` and `uno_set_constraint_upper_bound()` to:
  - [x] C API
  - [x] Julia interface
  - [x] Python interface
  - [x] Fortran interface

cc @amontoison @david0oo

Closes https://github.com/cvanaret/Uno/issues/671

### DIFF
--- a/interfaces/Julia/src/C_wrapper.jl
+++ b/interfaces/Julia/src/C_wrapper.jl
@@ -260,30 +260,21 @@ function uno_model(
   )
 end
 
-function uno_set_variables_bounds(model::Model, variables_lower_bounds::Vector{Float64}, variables_upper_bounds::Vector{Float64})
-  @assert model.nvar == length(variables_lower_bounds) == length(variables_upper_bounds)
+function uno_set_variables_lower_bounds(model::Model, variables_lower_bounds::Vector{Float64})
+  @assert model.nvar == length(variables_lower_bounds)
   GC.@preserve model begin
-    flag = uno_set_variables_bounds(model.c_model, variables_lower_bounds, variables_upper_bounds)
+    flag = uno_set_variables_lower_bounds(model.c_model, variables_lower_bounds)
   end
-  flag || error("Failed to set variables bounds via uno_set_variables_bounds.")
+  flag || error("Failed to set variables lower bounds via uno_set_variables_lower_bounds.")
   return
 end
 
-function uno_set_variable_lower_bound(model::Model, variable_index::Int, lower_bound::Float64)
-  @assert 1 <= variable_index <= model.nvar
+function uno_set_variables_upper_bounds(model::Model, variables_upper_bounds::Vector{Float64})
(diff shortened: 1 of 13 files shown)


interfaces/Julia/src/C_wrapper.jl
```diff
@@ -269,6 +269,24 @@ function uno_set_variables_bounds(model::Model, variables_lower_bounds::Vector{F
   return
 end
 
+function uno_set_variable_lower_bound(model::Model, variable_index::Int, lower_bound::Float64)
+  @assert 1 <= variable_index <= model.nvar
+  GC.@preserve model begin
+    flag = uno_set_variable_lower_bound(model.c_model, variable_index, lower_bound)
```
amontoison: `variable_index-1`

amontoison: I didn't interface these routines in `C_wrapper.jl` because they are based on C-indexing.

amontoison:
@cvanaret You updated the `README.md` of the Python interface.
We should do the same for C and Fortran.

cvanaret:
@amontoison the C and Fortran README are now up to date.

amontoison (changes requested):
You need the `model` as first argument for the new functions in the C / Fortran README.
It is not like Python 🐍 .

amontoison (approved):

PR #688, opened 2026-04-27
[APIs] Handle base indexing internally for `uno_[get|set]_*` functions with index parameter

cvanaret:
So far:
In Uno functions such as `uno_get_primal_solution_component(solver, index)`, the index should be in [0, upper_bound) (C style).

Now:
The index can be passed in the native base indexing (e.g., from 1 in Julia) and will be handled internally by Uno based on the model's `base_indexing` attribute.


cvanaret:
@amontoison we're good to go.

PR #474, opened 2025-12-13
[UnoSolver.jl] Reuse existing data structures from MOI on re-solve

amontoison:
close #419 

@franckgaga May I ask you to test the PR on your problem before that I merge it?

PR #689, opened 2026-04-28
[UnoSolver.jl] Update the MOI interface for new variable/constraint bounds

cvanaret:
cc @franckgaga @odow @amontoison


franckgaga:
Hello @cvanaret, what's the new variable/constraint bounds, something that changed internally in MOI ?

cvanaret:
@franckgaga
In https://github.com/cvanaret/Uno/pull/675, we added functions to set variable and constraints bounds on a model. In this PR (which extends https://github.com/cvanaret/Uno/pull/474), the C model may be updated using these functions, instead of rebuilt from scratch.

In the next few weeks, we will adapt Uno for re-optimization by recycling the internal workspaces.

franckgaga:
Wow that's great news! I'll re-run my benchmarks to see the improvements

PR #715, opened 2026-05-18
[APIs] handle `uno_default.opt` file properly

cvanaret:
- added `uno_default.opt file` in root directory
- preset may be set in option file with usual syntax
- handled it properly in AMPL interface
- improved the AMPL/C/Fortran READMEs

cc @nimgould


nimgould:
Very good, Charlie. Two questions, though

1.  I see that a line starting # is a comment. Does a # further along a line indicate "skip anything beyond the #"? It would be nice to allow inline comments
2. I don't see immediately a keyword to describe the actual solver used. There is a function uno_set_solver_preset( solver, "filtersqp" ) that allows one to specify (e.g.) filtersqp, but is that possible from the opt file?

nimgould:
Or is the intention just to use the ingredient keywords to mix things up?

cvanaret:
> Very good, Charlie. Two questions, though
> 
>     1. I see that a line starting # is a comment. Does a # further along a line indicate "skip anything beyond the #"? It would be nice to allow inline comments
> 
>     2. I don't see immediately a keyword to describe the actual solver used. There is a function uno_set_solver_preset( solver, "filtersqp" ) that allows one to specify (e.g.) filtersqp, but is that possible from the opt file?

Hi Nick,
- inline comments are allowed. I will add it explicitly to the README
- I left out `preset` on purpose, because it encompasses a whole set of options. You should set the preset (`uno_set_solver_preset`) and the option file (`uno_load_solver_option_file`) separately (in case of overlap, the options that are set last overwrite the previous ones).

nimgould:
Thank you. Good for 1. How should the name of the solver be passed into the CUTEst interface in 2? Is there a separate file that holds this name? Should I provide such a file? I presume the intention is not to recompute the CUTEst interface for each solver.

cvanaret:
> Thank you. Good for 1. How should the name of the solver be passed into the CUTEst interface in 2? Is there a separate file that holds this name? Should I provide such a file? I presume the intention is not to recompute the CUTEst interface for each solver.

Oh I hadn't thought of that. We don't want the user to recompile every time they want to change the preset. Perhaps the easiest way would be for me to provide both `uno_filtersqp.opt` and `uno_ipopt.opt`?

nimgould:
Do you really need two files? What will happen when Uno supports ten solvers? Are the parts that are specific to filtersqp the ones that have a filter_ suffix. If so, presumably an ipopt user will simply ignore them. How does reading the .opt file work? Does it start with all variables having defaults, and then anything that is not commented in the opt file replaces the default value? 

It would seem to me that you might have a uno_default.opt file that has a variable that names the solver, say

solver fitersqp

Having read this file, and determined the name of solver, it then looks for a second file uno_"solver".opt, where solver is the above name. If it finds this file, it reads this to override any relevant definitions from uno_default.opt

amontoison:
I asked Charlie a long time ago to support to pass the preset as any option and internally `uno_set_option` calls `uno_set_preset`.

I suppose that `uno_load_solver_option_file` is calling `uno_set_option` on each option parsed from the file so if the preset is always the first option in the option file, it should not be an issue.

cvanaret:
I now realize it's just easier to make `preset` a normal option :) I'll ping you when the changes are made so that you can have a look.

cvanaret:
Hi @nimgould, I've made changes so that the user can set the preset (= the solver) in the option file (usual syntax `preset filtersqp`). The options are set in the order in which they are defined in the file.
I also fixed the parser (some comments weren't properly interpreted).

nimgould:
Excellent, @cvanaret . I presume there is a function I call to set these values. Can I then see what values have been assigned, so that I can assign some of the values to fortran variables that I then use in CUTEst's output?

cvanaret:
Yes, you should call the function `uno_load_solver_option_file` (see [here](https://github.com/cvanaret/Uno/blob/f9e3db30c11201e65fe006db6c90efb9981f08f7/interfaces/Fortran/uno_fortran.f90#L260)).

At the moment, it is possible to ask the solver for the value of a particular option using `uno_get_solver_integer_option` (see [here](https://github.com/cvanaret/Uno/blob/f9e3db30c11201e65fe006db6c90efb9981f08f7/interfaces/Fortran/uno_fortran.f90#L322)), `uno_get_solver_double_option`, `uno_get_solver_bool_option`, and `uno_get_solver_string_option`.
The options are listed [here](https://github.com/cvanaret/Uno/blob/main/docs/options.md). Do you need to obtain this list programmatically?

nimgould:
No, I simply want (for example) to call uno_get_solver_string_option to get the name of the preset, so that I can write this to the one-line output summary that CUTEst interfaces traditionally produce. So I think we are ok. I will experiment with the fortran when I get a chance.

cvanaret:
Got it. Note that the user doesn't have to use a preset, they can assemble the solver themself using `constraint_relaxation_strategy`, `inequality_handling_method`, `globalization_mechanism`, and so on.

What may be of interest to you is the method description generated by Uno (such as `TR Fletcher-filter restoration inequality-constrained SQP method with exact Hessian and no inertia correction`). There is no function in the API yet, but the implementation effort is low (https://github.com/cvanaret/Uno/issues/421).

nimgould:
Yes, I understand. It would be quite helpful if Uno produced a single string to summarise the method built, although I appreciate that this may be tricky. The world famous FfricsqpmweHanic method you describe above needs a bit of polishing - it is shame that Mike Powell is no longer with us, he made a habit of producing great software with unpronounceable names

nimgould:
This looks good to me, I presume one of you adminers will merge this so that I can play

PR #751, opened 2026-05-28
[APIs] Added `get_solution_constraints` function

cvanaret:
Added:
- `uno_get_solution_constraints` function in C, Julia and Fortran
- `solution_constraints` read-only property in Python

cc @nimgould


nimgould:
That is great, I'll pop it in the CUTEst interface right now

PR #763, opened 2026-06-07
Add option utilities to C interface

stefphd:
Hi Charlie,
I have added a couple of utility functions in the C inteface. These may be usefull for iterating over all options.
I also noticed that in `uno_get_solver_option_type(void* solver, const char* option_name)` the `solver` input may be also removed, as the option type may be obtained directily from `Options::option_types`. For now, I kept the that function as is, since it may break examples or the other interfaces.

### DIFF
--- a/interfaces/C/README.md
+++ b/interfaces/C/README.md
@@ -129,13 +129,6 @@ uno_set_solver_bool_option(solver, "print_solution", true);
 uno_set_solver_string_option(solver, "hessian_model", "exact");
 ```
 
-To get the type, name, and index of a Uno option:
-```c
-uno_get_solver_option_type(solver, "max_iterations");
-uno_int index = uno_get_solver_option_index("max_iterations"); // -1 if not found
-const char* name = uno_get_solver_option_name(0); // NULL if not found
-```
-
 Options can also be loaded from a file that contains an `option value` pair per line (the example file `uno_default.opt` in the root directory contains Uno's default options). The options are set in this particular order. Anything that follows a `#` is treated as a comment and is ignored.
 ```c
 uno_load_solver_option_file(solver, "uno.opt");
@@ -153,6 +146,32 @@ const char* uno_get_solver_string_option(solver, "hessian_model");
 Setting a preset has Uno mimic an existing solver:
 ```c
 uno_set_solver_preset(solver, "filtersqp");
+```
+
+Getting the type of an option from the Uno solver:
+```c
+uno_int uno_get_solver_option_type(solver, "max_iterations");
+```
+
+Iterating over the Uno options:
+```c
+uno_option_iterator uno_option_begin_iterator();
+uno_option_iterator uno_option_end_iterator();
+void uno_option_next_iterator(&option_iterator);
+const char* uno_option_iterator_name(option_iterator);
+uno_int uno_option_iterator_type(option_iterator);
+void uno_option_destroy_iterator(&option_iterator);
+```
+
+The iterator is automatically destroyed when the end is reached in `uno_option_next_iterator`. After destruction, the iterator is set to null.
+
+Example of iterating over Uno options:
+```c
+for (uno_option_iterator it = uno_option_begin_iterator(); it != uno_option_end_iterator(); uno_option_next_iterator(&it)) {
+   const char* name = uno_option_iterator_name(it);
+   uno_int type = uno_option_iterator_type(it);
+   // user's code ...
+}
 ```
 
 ## Setting solver callbacks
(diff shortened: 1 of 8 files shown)


cvanaret:
Thanks @stefphd! I'm not sure why you need the order of the options though. The order in [Options.cpp](https://github.com/cvanaret/Uno/blob/main/uno/options/Options.cpp) is a bit random.
You're right about `uno_get_solver_option_type`. Let's keep it that way for now, we'll change it if we introduce more breaking changes.

stefphd:
Hi Charlie,

I noticed that the options are returned in no particular order. However, in the C interface it may still be necessary for users to iterate over all available options (as in my case). Given the lack of native iterator support in C, the only workaround I can think of is something along these lines:

```c
uno_int it = 0;
while (true) {
   const char* option_name = uno_get_solver_option_name(it);
   if (option_name == NULL) { // null when out of range
      break;
   }
   uno_int option_type = uno_get_solver_option_type(solver, option_name);
   // [..]
   it++;
}
```
Let me know if you see a cleaner approach in the C API.

stefphd:
We could also provide `uno_int uno_get_solver_option_count()` so that users can iterate over available solver options in a more idiomatic way

```c
uno_int option_count = uno_get_solver_option_count();
for (uno_int i = 0; i < option_count; ++i) {
    const char* option_name = uno_get_solver_option_name(i);
    uno_int option_type = uno_get_solver_option_type(solver, option_name);
    // user code here
}
```

cvanaret:
Hi @stefphd,
@amontoison and I think that using an iterator would be more robust, something like:

```c
for (option_iterator it = create_option_iterator(); it != end_option_iterator(); next_option_iterator(&it)) {
    printf("%s -> %d\n", option_iterator_key(it), option_iterator_value(it));
}
```
See [Claude's implementation](https://claude.ai/share/f49b8501-f3c3-4242-b28c-e22e35c8ef4b). What do you think?

stefphd:
This is actually the other implementation Claude also suggested to me ;)
I’m fine with both. Using an iterator is of course more robust.

stefphd:
I've made an implemetation using the iterator, with just few differences from the Claude implementation.

- `uno_option_iterator` wraps the C++ Option iterator
- iterator is null and freed when end is reached
- I noted that the bool option `write_solution_to_file` was missing in DefaultOptions, although present in options.md. I've added the option in Default Options
- The string options `preset` and `option_file` need special care when the getter is used, as it may not be in the string option list

I will upload the readme in a couple of days as well.

cvanaret:
@stefphd it looks great.

I suggest that we merge it as is (when the README is updated) and that we try Claude's suggestion later on (that'd be nice to have but not critical):
> If you're willing to change the call site to something like while (!iterator_at_end(it)) or iterators_equal(it, end), then iterator can be a by-value struct holding the C++ iterator inline, with zero allocation

stefphd:
I’ve just uploaded the README, plus few other minor changes.

I also noticed that `option_file` was not handled in `uno_set_solver_string_option`; it is now handled in the same way as `preset`.

One other thing: I noticed that exceptions (e.g. `std::runtime_error)` are thrown in the C interface. I’m not sure this is ideal, since C has no exception handling and ABI may not match between different compilers. What do you think? (Of course, this would be a separate PR)

cvanaret:
> I also noticed that `option_file` was not handled in `uno_set_solver_string_option`; it is now handled in the same way as `preset`.

Thanks!

> One other thing: I noticed that exceptions (e.g. `std::runtime_error)` are thrown in the C interface. I’m not sure this is ideal, since C has no exception handling and ABI may not match between different compilers. What do you think? (Of course, this would be a separate PR)

Good point. I constantly get scolded by the compiler because the C functions throw. I guess the only option is to return a bool and to set the return to an out parameter, but boy is this ugly.

stefphd:
> Good point. I constantly get scolded by the compiler because the C functions throw. I guess the only option is to return a bool and to set the return to an out parameter, but boy is this ugly.

Yes, that's ugly, although this approach is used in other software (e.g. WORHP). A lower-impact option could be to add a `const char* uno_get_last_error()` function that users can call after API calls.

```c
// API call, e.g. get option, solve
// ...
const char* errmsg = uno_get_last_error();
if (errmsg) {
    fprintf(stderr, "%s", errmsg);
}
```

cvanaret:
> Yes, that's ugly, although this approach is used in other software (e.g. WORHP). A lower-impact option could be to add a `const char* uno_get_last_error()` function that users can call after API calls.

OK I think returning a bool is still less ugly than the `uno_get_last_error` option. Let's do that separately (we'll have to break the API).
Is the PR ready for review?

stefphd:
Yes, the PR is ready for review.

cvanaret (approved):

cvanaret:
All good for me. I will make a few tiny changes in a separate PR (like renaming things). Thanks a lot!

PR #784, opened 2026-06-18
Added `IQPSolver` abstraction on top of `LPSolver` and `QPSolver`

cvanaret:


cvanaret:
@amontoison we're good to go!

PR #797, opened 2026-06-29
Automatic preset

cvanaret:
- Implemented an `auto` preset. The oracle decides between `ipopt` and `filtersqp` based on the dimensions of the model.
- Added two checks (SQP and interior-point) in `check_auto_preset.jl`


cvanaret:
@amontoison it seems to be working :)
What I would need from you (when you have 5 minutes) would be to expose [`uno_get_method_description`](https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/src/libuno.jl#L260) in MOI, so that I can test the method used by Uno in the [new workflow](https://github.com/cvanaret/Uno/actions/runs/28384857330/job/84098517460?pr=797).

PR #771, opened 2026-06-11
Dynamic loading for HSL

jgillis:
Hi Charlie!

Finally found some time to get Uno into casadi..

I did some minor modifications to get it more in line with expected behavior from ipopt.
I'm happy to just roll with casadi-patches against your official releases, but you may also chose to merge some of this.

PR #808, opened 2026-07-08
Standalone dynamic loader (follow-up of #771)

cvanaret:
Small refactoring to make the dynamic loader standalone (independent of HSL), so that it can be reused for loading other libraries (like libkrylov).

cc @amontoison @jgillis

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,8 +132,8 @@ endfunction()
 # ======================
 
 # HSL (MA27/MA57)
-# IPOPT-style runtime loading: don't link libhsl, dlopen it on first use and
-# resolve MA27/MA57 by symbol name (see uno/ingredients/subproblem_solvers/HSL).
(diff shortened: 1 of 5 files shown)


cvanaret:
I moved most of the dynamic loading code to `tools/DynamicLoader.cpp`.
I'm not sure about the content of `HSLLoader::load_hsl_library` though (see [here](https://github.com/cvanaret/Uno/pull/808/changes#diff-8249a849dfa0dba0d0b12c2257c9e9fc2870280ee0a59218c3d1418d568af38cR34)). Should we leave it as it and let each solver handle its own handle caching?

amontoison:
We should keep `load_hsl_library` in `HSLLoader.cpp` because it relies on `UNO_HSL_DEFAULT_LIBRARY`.

For dynamically loaded dependency (`libhsl` or `libkrylov`), we will have a different handle.

PR #812, opened 2026-07-09
Second-order corrections for line-search IPM

cvanaret:
Second-order corrections implemented for line-search interior-point methods


cvanaret:
OK, it seems to work. I'll work on improving the memory allocations when I'm back from my vacation.

PR #818, opened 2026-07-30
Added flush calls to write the opt logger info to the disk immediately.

friedenhe:
Problem: The current unopy logger file writes the optimization progress to the disk ONLY after the optimization is fully completed. This is not a problem if the optimization is fast. But for CFD-based optimization, each optimization iteration may take tens of minutes, so it is very useful to monitor the progress while the optimization is still running.

Fix: Add a few Logger:flush() calls to make the logger write the optimization progress to the disk immediately.

### DIFF
--- a/uno/ingredients/constraint_relaxation_strategies/FeasibilityRestoration.cpp
+++ b/uno/ingredients/constraint_relaxation_strategies/FeasibilityRestoration.cpp
@@ -253,22 +253,27 @@ namespace uno {
          warmstart_information.no_changes();
       }
 
-      // check termination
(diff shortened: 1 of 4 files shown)


friedenhe:
It seems that there is an internet error in the test that is not related to the code change.

cvanaret:
Hi @friedenhe, good catch! Did you close the PR on purpose?

friedenhe:
@cvanaret Thanks for checking. Yes, I accidentally pushed some additional changes to this PR. I closed this old PR and submitted a new one for the logger::push feature.

PR #820, opened 2026-08-03
[CI] Parameterized Julia equality-constrained tests with linear solver

cvanaret:
- replaced `runtests_uno_ipopt_[ma27|ma57|ma86|mumps|ssids].jl` with single `runtests_uno_ipopt.jl` parameterized with command line argument
- also parameterized all the other tests for equality-constrained problems
- adapted the workflows

I'll need more self-hosted runners...


cvanaret:
Apparently, the few failures are due to small differences in behavior between linear solvers, and the constraint violation cannot be sufficiently decreased.
I assume this would vanish with iterative refinement. We could also ignore it because it's with an identity Hessian anyway.
Edit: commented out the tests with identity Hessian.

PR #830, opened 2026-08-07
[UnoSolver.jl] Update README.md

amontoison:
The "BLAS and LAPACK" section lists MKL, AOCL and Apple Accelerate as alternative backends but doesn't say which platform each one targets. This adds a short hint so users can quickly pick the backend suited to their CPU.

`AOCL.jl` was registered this morning: https://github.com/JuliaLinearAlgebra/AOCL.jl

### DIFF
--- a/interfaces/Julia/README.md
+++ b/interfaces/Julia/README.md
@@ -158,7 +158,8 @@ set_attribute(model, "LP_solver", "HiGHS")
 
 ## BLAS and LAPACK demuxer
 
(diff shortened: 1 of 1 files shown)


interfaces/Julia/README.md
```diff
@@ -193,11 +194,21 @@ using MKL
...
 ```
 
+### AOCL
+
+If you have [AOCL.jl](https://github.com/JuliaLinearAlgebra/AOCL.jl) installed,
+switch to AOCL by adding `using AOCL` to your code:
+
+```julia
+using AOCL
+using Ipopt
```
cvanaret: Copy-paste? 😛

amontoison:
```suggestion
using UnoSolver
```

amontoison: Good catch!

cvanaret (approved):

PR #851, opened 2026-08-15
Tentative upgrade to ASL_jll v0.1.5

cvanaret:
Tentative upgrade to latest ASL. I'm suspecting that ASL_jll.jl is not up to date, let's give it a try.

What I'm after:
- Sphese and Hvcompe: Hessian evaluation and Hessian-vector product with an extra nerror parameter
- (minor) const correctness of the ASL signatures

AFAICT there were no breaking changes in the API, only additional functions.

### DIFF
--- a/.github/julia/build_tarballs_release.jl
+++ b/.github/julia/build_tarballs_release.jl
@@ -229,7 +229,7 @@ dependencies = [
     BuildDependency(PackageSpec(name="BQPD_jll", uuid="1325ac01-0a49-589f-8355-43321054aaab")),
     Dependency(PackageSpec(name="HSL_jll", uuid="017b0a0e-03f4-516a-9b91-836bbd1904dd")),
     Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70")),
-    Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.3"),
+    Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.5"),
     Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2")),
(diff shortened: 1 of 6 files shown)


cvanaret:
That's what I thought. The latest changes to https://www.netlib.org/ampl/solvers/ are not in ASL_jll.jl.
@amontoison do I simply need to change the tgz hash and bump the version in https://github.com/JuliaPackaging/Yggdrasil/blob/master/A/ASL/build_tarballs.jl?

amontoison:
@cvanaret It is a little bit more complex for `ASL_jll.jl` because it is used by many artifacts and I don't want to break anything.
What I propose is to update the General registry with a PR to add a compat entry on `ASL_jll.jl` for JLL compiled without it.
The concerned JLL are:
- SHOT
- Ipopt
- IpoptMKL
- Bonmin
- Cbc
- Couenne

The only JLL with a compat entry on `ASL_jll.jl` is... Uno!

Once it is done, we can update all the related `build_tarballs.jl` files to add the compat entry but not trigger a cross-compilation (no CI).

Once it is done, we can use the GitHub repository of ASL (like you did for ASL2) and specify a meaningful version (XXXX.YY.ZZ).
It will be great because we don't know what version was used for `ASL_jll`, AMPL recycles the same link for new `solvers.tgz`...

I broke a bunch of packages with the upgrade SCOTCH 6 --> SCOTCH 7 in the past and I don't want to reproduce it with the upgrade of ASL:
- https://github.com/JuliaRegistries/General/pull/81580
- https://github.com/JuliaPackaging/Yggdrasil/issues/6569

What do you think Oscar?
cc @odow

odow:
I've never wanted to touch ASL_jll.jl for this reason...

> AFAICT there were no breaking changes in the API, only additional functions.

Are you reeeeeeally sure?

cvanaret:
> Are you reeeeeeally sure?

Not a 100% :)

I think the only way out is to have Uno depend on ASL2(_jll), which can be kept in sync.

amontoison:
@cvanaret Can you provide a draft PR on Yggdrasil to update `ASL_jll.jl` such that we can cross-compile it on a fork like we do for `UnoUtils_jll.jl` ?

It will give us a chance to test it and check if it breaks a library.
If we have all the tests of `AmplNLWriter.jl` and `AmplNLReader.jl` working on linux / mac / windows, maybe we can upgrade it.

PR #866, opened 2026-08-20
Function scaling

cvanaret:
Function scaling a la IPOPT: compute the gradients at the initial iterate and use their magnitude to scale the functions.


cvanaret:
@amontoison that's what I'm talking about :muscle:

PR #874, opened 2026-08-23
Detect diverging iterates (added status DIVERGING_ITERATE)

cvanaret:
Detect diverging iterates when their inf norm exceeds 1e20 (added status DIVERGING_ITERATE)


cvanaret:
@amontoison got saved twice by your option checking script :laughing:

PR #886, opened 2026-08-28
Fixed IPM bugs

cvanaret:
- **big one**: line search: forgot to scale step length with fraction-to-boundary step length when calling `this->constraint_relaxation_strategy->is_iterate_acceptable`. This would mess up the predicted reduction (it would appear negative for the unconstrained feasibility problem... :shaking_face: )
- **small one**: when switching to feasibility restoration, push the slacks into the interior of their bounds


amontoison:
😛 😛 😛 
unacceptable

PR #889, opened 2026-08-28
Attempt: do not compile HiGHS on the fly for MinGW

cvanaret:
Take the precompiled `libhighs` from UnoUtils instead of compiling HiGHS on the fly for MinGW

Worth a try 😅


amontoison:
Is it not possible to use directly `highspy` as a dependency of `unopy` on all platforms?

cvanaret:
Still TLS mismatch with the HiGHS C API...

PR #896, opened 2026-08-30
Enable MC64 scaling in MA57 for Mittelmann benchmark

cvanaret:
MA57's MC64 scaling is expensive, but the barrier solver becomes more robust over the whole benchmark.


cvanaret:
@amontoison working well :)

cvanaret:
Looking great! 43/47 with 10min/instance.
Next steps:
- [x] add an option for MC64
- [x] write an option file for the Mittelmann benchmark

PR #902, opened 2026-08-31
IPM fixes

cvanaret:
- Use `HomogeneousEqualityConstrained` reformulation even if no inequalities/bounds: the restoration phase will thank you!
- Cap the LS step length *including the FTB scaling* to `minimum_step_length`
- Set bound duals to +/-1 when switching back. **TODO**: compute their values using the linearized complementarity equation
- Take the prox coefficient from the *feasibility* barrier method (the optimality inequality handling method may be `NoInequalityReformulation`)
- When switching back, reset the duals of the *trial iterate* (not the current iterate!)

### DIFF
--- a/uno/Uno.cpp
+++ b/uno/Uno.cpp
@@ -48,8 +48,7 @@ namespace uno {
       // reformulate the model if:
       // - the user wants an interior-point method with log barrier function
       // - the model has bound constraints or inequality constraints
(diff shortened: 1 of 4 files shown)


cvanaret:
Regression looks bad 😄 But these are legit bug fixes. More to come!

PR #903, opened 2026-09-01
Use a filter method for globalizing the ipopt restoration phase

cvanaret:
- use a filter method for globalizing the `ipopt` restoration phase
- properly define the l1 relaxation progress measures and local models
- properly initialize the elastics for SQP restoration
- [WIP] use the same `is_infeasibility_sufficiently_reduced` in `FletcherFilterMethod` as in `WaechterFilterMethod`: require 10% of decrease compared to infeasibility at reference point + filter acceptability.

This should have a **large** impact on robustness.


cvanaret:
- ipopt + SSIDS fails on bfpc, which is an unrelated issue
- MINLPTests' 105_013 instance was reported as "limit iteration" and commented out (ok for SLP)

PR #906, opened 2026-09-03
Relax linear constraints as well in `l1RelaxedProblem`

cvanaret:
The idea *not* to relax the linear constraints in `l1RelaxedProblem` in the presence of a trust-region constraint deserves to be investigated, but for the elastic initialization of the IPM, this messes things up completely (if the current point is infeasible wrt the linear constraints, there's no elastics to catch this)


cvanaret:
We gained on `qcqp500-3nc` but lost on the other QCQPs (they're probably much larger because of the elastics). It's fine, now the failures time out. Let's see what happens in the official benchmark.

PR #914, opened 2026-09-05
Add slacks in the barrier reformulation, not in the model

cvanaret:
TODO:
- [x] moved slacks from `HomogeneousEqualityConstrainedModel` to `PrimalDualInteriorPointProblem`
- [x] phase transitions
- [x] correct residuals (stat and cmp from slacked model, constraint violation from original model)
- [x] residual scaling
- [ ] fix remaining failures on infeasible instances

Consequences of the refactoring:
- keep track of structure in a finer way (identity block in inequality Jacobian, regularize only original variables)
- it's cheaper (fewer variables) to do l1 relaxation then slacks, than slacks then l1 relaxation


cvanaret:
Superseded by #936

PR #930, opened 2026-09-13
[WIP] Use OpenBLAS for the packaging

cvanaret:
cc @amontoison 

Bundle OpenBLAS for:
- precompiled archives
- Python wheels


cvanaret:
Mh looks like the ~~macOS and Windows wheels segfault~~ the macOS wheel segfaults and the Windows wheel doesn't find libhighs_extras_openblas (makes sense because HiGHS is compiled on the fly)

cvanaret:
Windows wheels fixed :v:

cvanaret:
Back to reference BLAS/LAPACK for the macos-15-intel wheel.

cvanaret:
macos-15-intel wheel shipped with reference BLAS and LAPACK for now.

Claude's diagnosis:
Build OpenBLAS shared on `x86_64-apple-darwin` (`NO_STATIC=1`, drop `NO_SHARED=1` for that target only), matching Julia's working `OpenBLAS32_jll`. The dylib link resolves the mangled locals internally, and delocate bundles the `.dylib` into the wheel. This is the change that directly targets the static-vs-shared root cause — the one structural difference between your build and the known-good one that we hadn't touched. Highest chance of making OpenBLAS actually work on Intel.

PR #935, opened 2026-09-16
Bug fix in `PrimalDualInteriorPointProblem::create_initial_iterate`

cvanaret:
Zero the slacks in `PrimalDualInteriorPointProblem::create_initial_iterate` to get an accurate evaluation of the *unslacked* constraints


cvanaret:
We're winning on `NARX_CFy` but losing on the `robot`s... To be continued!

